PR #2, opened 2015-02-19
Implemented several core classes.

tomrijnbeek:


src/Core/Environment.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// <summary>
+    /// Collection of platform-specific functions.
+    /// </summary>
+    public static class Environment
+    {
+        /// <summary>
+        /// The platform the application is currently running on.
+        /// </summary>
+        public static Platform CurrentPlatform
```
amulware: This might be nicer with caching. At least I can see no way in which the Platform can ever change at runtime.

PR #3, opened 2015-02-19
Added input manager.

tomrijnbeek:


amulware:
Having the inputmanater be a static class is somewhat ugly. However, making it a Singleton would only add annoying '.Instance' to every call, and making it a regular class does not seem sensible either.

Should possibly be re-evaluated at some point in the future.

PR #5, opened 2015-02-20
Added Hungarian algorithm.

tomrijnbeek:


src/Algorithms/HungarianAlgorithm.cs
```diff
@@ -0,0 +1,387 @@
...
+
+        /// <summary>
+        /// Evaluates a minimal cost matching on two-dimensional vectors using the least-squares metric.
+        /// </summary>
+        /// <param name="from">The source vectors.</param>
+        /// <param name="to">The destination vectors.</param>
+        /// <returns>The minimal cost matching based on the least-squares metric. A value of -1 means the source vector is not assigned.</returns>
+        public static int[] Run(Vector2[] from, Vector2[] to)
+        {
+            return HungarianAlgorithm.Run(from, to, (u, v) => (v - u).LengthSquared);
```
amulware: I'd be curious to see how much overhead this lambda adds, compared to writing out the loop directly.

Since this will likely be one of the most common usages of the algorithm, it may be worth testing this, and possibly writing out the loop from above.

src/Algorithms/HungarianAlgorithm.cs
```diff
@@ -0,0 +1,387 @@
...
+            this.labelSources = new float[this.dim];
+            this.labelDests = new float[this.dim];
+            this.minSlackDestBySource = new int[this.dim];
+            this.minSlackValueBySource = new float[this.dim];
+            this.matchedSources = new bool[this.dim];
+            this.parentSourceByCommittedDest = new int[this.dim];
+            this.sourceMatches = new int[this.dim];
+            this.destMatches = new int[this.dim];
+
+            for (int i = 0; i < this.dim; i++)
```
amulware: This loop could be skipped by storing indices+1 instead, and leaving 0 as uninitialised value.

The Array.Copy in line 134 could then be combined with the loop in line 135, for a single instead of two traversals.

amulware:
I'd like to see https://github.com/beardgame/utilities/pull/5#discussion_r25121171 addressed before this is merged.
Possibly also https://github.com/beardgame/utilities/pull/5#discussion_r25121154, though I find that less critical.

Otherwise looking good!

src/Algorithms/HungarianAlgorithm.cs
tomrijnbeek: This loop is insignificant compared to the overall running time of the algorithm (O(n) vs O(n^3)). Furthermore this would make the usage of these arrays less intuitive and very prone to errors in future changes. Therefore I think there is no valid argument to change this.

amulware: Fair enough. Nevermind this then.

PR #7, opened 2015-02-20
Added several geometry related classes.

tomrijnbeek:


src/Math/Geometry/PolarPosition.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        #region IEquatable implementation
+        /// <summary>
+        /// Indicates whether the current object is equal to another object of the same type.
+        /// </summary>
+        /// <returns>
+        /// true if the current object is equal to the <paramref name="other"/> parameter; otherwise, false.
+        /// </returns>
+        /// <param name="other">An object to compare with this object.</param>
+        public bool Equals(PolarPosition other)
```
amulware: Consider also overriding == and != as well as Equals(object) and GetHashCode().

However, be sure to have a special case for hashing in case of r == 0 (I suggest simply returning hash code 0 for that case and letting Resharper generate the rest).

src/Math/Geometry/Rectangle.cs
```diff
@@ -0,0 +1,162 @@
...
+        /// <param name="height">Height of the rectangle.</param>
+        public Rectangle(Vector2 xy, float width, float height)
+            : this(xy.X, xy.Y, width, height) { }
+
+        /// <summary>
+        /// Creates a new rectangle.
+        /// </summary>
+        /// <param name="upperLeft">Coordinates of the upper left corner.</param>
+        /// <param name="bottomRight">Coordinates of the bottom right corner.</param>
+        public Rectangle(Vector2 upperLeft, Vector2 bottomRight)
```
amulware: This may be confused with a possible Rectangle(Vector2 upperLeft, Vector2 size).

Consider using static methods for these constructurs instead?

src/Math/Geometry/Rectangle.cs
```diff
@@ -0,0 +1,162 @@
...
+        /// </summary>
+        /// <param name="x">X-coordinate of the upper left corner.</param>
+        /// <param name="y">Y-coordinate of the upper left corner.</param>
+        /// <param name="width">Width of the rectangle.</param>
+        /// <param name="height">Height of the rectangle.</param>
+        public Rectangle(float x, float y, float width, float height)
+        {
+            this.x = x;
+            this.y = y;
+            this.width = width;
```
amulware: This should enforce non-negative width and height.

amulware:
I think my comments should be addressed before this is merged to make it more stable/reliable and improve ease of use.

src/Math/Geometry/Rectangle.cs
tomrijnbeek: I disagree. Vectors should never ever be used for size and the parameters have clear enough names I think.

amulware: I have seen vectors used for sizes all over the place, after all a size is just s relative vector, while a point is an absolute one. Also, while the parameters are named nicely, one usually does not see their names, when reading code that uses them, i.e. `var r = new Rectangle(new Vector2(1, 1), new Vector2(2, 2));`

It is also not very consistent with the other constructors, which DO require you to specify the size, and not the coordinates of the bottom left corner. In fact, next to the other two constructors, I would assume this one wants a size, if I did not know better.

tomrijnbeek:
Please revise.

amulware:
Looking great now! :)

PR #12, opened 2015-02-24
Various core types

amulware:
What it says on the tin.

### DIFF
--- a/src/Core/Box.cs
+++ b/src/Core/Box.cs
@@ -1,5 +1,4 @@
-﻿using System;
-
+﻿
 namespace Bearded.Utilities
 {
     /// <summary>
--- a/src/Core/Box.cs
+++ b/src/Core/Box.cs
@@ -7,7 +7,7 @@ namespace Bearded.Utilities
     public static class Box
     {
         /// <summary>
-        /// Gets a new box witha  given value.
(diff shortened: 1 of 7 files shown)


src/Core/Box.cs
```diff
@@ -0,0 +1,42 @@
+
+namespace Bearded.Utilities
+{
+    /// <summary>
+    /// Static Box class to help with box creation.
+    /// </summary>
+    public static class Box
+    {
+        /// <summary>
+        /// Gets a new box witha  given value.
```
tomrijnbeek: typo

src/Core/Box.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        /// <param name="value">The value of the box.</param>
+        public static Box<T> This<T>(T value)
```
tomrijnbeek: is this supposed to be an extension method?

amulware: Nope, intended usage is:

`var t = Box.This(DateTime.Now);`

Do you think I should turn it into an extension method (would be valid on all value types).
Or should I rename it to `From` or something else instead?

tomrijnbeek: I see. Is fine then, didn't understand :)

PR #16, opened 2015-02-24
Input actions

amulware:
This relies on the linq helpers and therefore on static randoms. Should be merged after those two.


src/Input/GamePadAction.cs
```diff
@@ -0,0 +1,381 @@
...
+            }
+
+        }
+
+        sealed private class GamePadAxisAction : IAction
+        {
+            private readonly static Dictionary<string, AxisSelector> axisSelectors = new Dictionary<string, AxisSelector>
+            {
+                { "+x", s => s.ThumbSticks.Left.X },
+                { "-x", s => -s.ThumbSticks.Left.X },
```
tomrijnbeek: should we consider renaming these to `+x1` for left and `+x2` for right stick. Or maybe `left +x` and `right +x`?

tomrijnbeek:
I think it might be interesting to also add MouseActions, as the input currently seems limited to keyboard and mouse only. Consider the use case for example where there are several control schemes where left, right and middle mouse buttons might switch functionality.

Not a necessity, just an interesting feature to consider in the future.

src/Input/GamePadAction.cs
amulware: Given that some controllers might not have two sticks, or might have the sticks reversed, or the x and y axis for some sticks reversed... I think in the least a lot more research and maybe a more flexible solution is needed.

amulware: In either case, these strings are not meant for being shown to the user, but only for (de)serialising where needed (safe control options mainly I suppose).

amulware:
Mouse actions would need some smart design. Buttons would be fairly trivial, but the cursor probably always has to be a special case since it behaves so differently, and uniquely.

Either way, idea filed away for future reference.

PR #17, opened 2015-02-24
Priority queue & unit tests

tomrijnbeek:
I assume the unit tests are sufficient enough to guarantee that the priority queues work correctly. I consider also making the static priority queue tests for the normal priority queue, but the existing tests pretty much contain all of the static tests as a subset.

### DIFF
--- a/src/Collections/PriorityQueue.cs
+++ b/src/Collections/PriorityQueue.cs
@@ -32,6 +32,7 @@ namespace Bearded.Utilities.Collections
         public PriorityQueue(IEnumerable<KeyValuePair<TPriority, TValue>> data)
         {
             this.data = data.ToArray();
+            this.Count = this.data.Length;
             for (int i = 0; i < this.data.Length; i++)
                 this.valueDict.Add(this.data[i].Value, i);
             for (int i = this.data.Length / 2 - 1; i >= 0; i--)
@@ -46,6 +47,8 @@ namespace Bearded.Utilities.Collections
         public void DecreasePriority(TValue value, TPriority newPriority)
         {
             int i = this.valueDict[value];
+            if (this.data[i].Key.CompareTo(newPriority) == -1)
+                throw new InvalidOperationException("Can not increase the priority.");
             this.data[i] = new KeyValuePair<TPriority, TValue>(newPriority, value);
             this.cascadeUp(i);
         }
--- a/src/Collections/PriorityQueue.cs
+++ b/src/Collections/PriorityQueue.cs
@@ -1,0 +1,97 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Bearded.Utilities.Collections
+{
+    /// <summary>
+    /// A priority queue that is implemented with a binary min-heap.
+    /// </summary>
+    /// <typeparam name="TPriority"></typeparam>
+    /// <typeparam name="TValue"></typeparam>
+    public sealed class PriorityQueue<TPriority, TValue> : StaticPriorityQueue<TPriority, TValue> where TPriority : IComparable<TPriority>
+    {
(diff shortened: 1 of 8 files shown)


src/Collections/PriorityQueue.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        /// <summary>
+        /// Decreases the priority of the specified element [O(log n)].
+        /// </summary>
+        /// <param name="value">The element that should change.</param>
+        /// <param name="newPriority">The new priority of the element.</param>
+        public void DecreasePriority(TValue value, TPriority newPriority)
+        {
+            int i = this.valueDict[value];
+            if (this.data[i].Key.CompareTo(newPriority) == -1)
```
amulware: This should be replaced by a `< 0` check.
CompareTo is not required to return -1/1. The documentation specifies only that the returned value has to be negative or positive.

src/Collections/PriorityQueue.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        /// <summary>
+        /// Adds a new element to the end of the tree.
+        /// </summary>
+        /// <param name="priority">The priority of the new element.</param>
+        /// <param name="value">The element itself.</param>
+        protected override void add(TPriority priority, TValue value)
+        {
+            this.valueDict.Add(value, this.Count);
```
amulware: This will throw a `ArgumentException` when adding a duplicate value. Is that intended behaviour?

src/Collections/StaticPriorityQueue.cs
```diff
@@ -0,0 +1,205 @@
...
+        /// Cascades the changes upwards in the tree starting from the specified index [O(log n)].
+        /// </summary>
+        /// <param name="i">The index at which the heap property might be violated.</param>
+        protected void cascadeUp(int i)
+        {
+            while (i > 0)
+            {
+                var parent = getParent(i);
+
+                if (data[i].Key.CompareTo(data[parent].Key) == -1)
```
amulware: Value of -1 is not guaranteed, replace with `< 0` check.

src/Collections/StaticPriorityQueue.cs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            while (true)
+            {
+                var left = getLeftChild(i);
+                var right = getRightChild(i);
+                var smallest = i;
+
+                if (left < this.Count && this.data[left].Key.CompareTo(this.data[smallest].Key) == -1)
+                    smallest = left;
+                if (right < this.Count && this.data[right].Key.CompareTo(this.data[smallest].Key) == -1)
```
amulware: Another two `< 0` checks needed here.

src/Collections/PriorityQueue.cs
tomrijnbeek: Yes

amulware:
Apart from the issues with CompareTo, this looks fine to me.
Once those are fixed, I will merge.

I do have some thoughts on how to make several of the operations a slight bit more efficient, but I will do some tests on that myself, when I find the time. (Mostly some constant time factors, or even additions by decreasing number of dictionary accesses.)

tomrijnbeek:
If performance is really a big problem, the user should use Fibonacci heaps instead. They are asymptotically faster than binary heaps, but due to a higher constant they are usually slower for practical cases. In these practical cases, the small improvements you speak of might already be unnecessary, and at the point they become significant, the user should probably switch to Fibonacci heaps if performance is indeed important.

amulware:
I think your approach is too theoretical. From a game development perspective, constant factors are pretty important. Problems tend to be small and many, so this adds up, and another 20-50% in a critical system may help the entire application.

But as I said, I'll take a look myself if I find the time.

PR #23, opened 2015-03-12
added file modified watcher

amulware:


src/IO/FileModifiedWatcher.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            this.lastModified = this.getLastWriteTime();
+        }
+
+        /// <summary>
+        /// Checks whether the file was changed since the last reset of the watcher.
+        /// </summary>
+        /// <param name="resetModified">Whether to reset the watcher after checking for changes.</param>
+        /// <returns>True, if the file has a different last-write time stamp than when the watcher was reset last.
+        /// True if the file was created or deleted since the last reset.
```
tomrijnbeek: Two times true? Looks a bit weird to me.

amulware: Want me to reword? The content is correct.

tomrijnbeek: Send a pull request if you want. It's a bit confusing that true is considered twice.

PR #26, opened 2015-04-04
added generic logging methods, resolves #25

### LINKED ISSUE #25: Add Log<T> method
It should be possible to log arbitrary objects (same as Console.WriteLine), which can trivially be implemented using generics.

amulware:


tomrijnbeek:
What is the reason for using generic methods as opposed to having an overload with an `object` parameter?

amulware:
To avoid boxing for value types.

PR #28, opened 2015-04-04
small optimisations for interpolation edge cases

amulware:


src/Math/Interpolate.cs
```diff
@@ -158,7 +188,12 @@ public static float Hermite(float value1, float tangent1, float value2, float ta
         /// <returns>The interpolated value.</returns>
         public static float SmoothStep(float from, float to, float t)
         {
-            return Interpolate.Hermite(from, 0, to, 0, t.Clamped(0, 1));
+            if (t <= 0)
```
tomrijnbeek: These checks only add more overhead, as the `Hermite` function already checks for these. Of course this would get rid of one extra call for these edge cases, but in the meanwhile adding more overhead for intermediate steps.

PR #32, opened 2015-04-06
added greedy rectangle bin packing algorithm and tests

amulware:
The code is a somewhat cleaned up version of the one used in the Roche Fusion sprite packing tool.

The algorithm itself is almost a 1:1 translation from Jake Gordon's javascript implementation here: https://github.com/jakesgordon/bin-packing/blob/master/js/packer.growing.js


tomrijnbeek:
I assume there will be no license problems?

amulware:
The original implementation is MIT, no problem there.

PR #35, opened 2015-04-08
Xunit

### LINKED ISSUE #33: Use different test framework.
While the current testing framework (MStest) is very convenient, it is only supported in Visual Studio. Hence the tests only work on Windows. While the library should be completely usable through Mono, we would have to switch to either [xUnit](https://github.com/xunit/xunit) or [NUnit](https://github.com/nunit/nunit) to make the library wholly usable among all platforms.

I personally think xUnit would be the better choice, as it seems to allow for integration with Visual Studio in a similar manner as MSBuild.

tomrijnbeek:
I replaced the MSTest tests by xUnit tests. I also added it to the travis configuration file, hence this pull request depends on #34.

Right now I can't get the Visual Studio runner to work, but the Resharper and console runner work just fine, so people are encouraged to use them.

Resolves #33.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: csharp
 solution: Bearded.Utilities.sln
 install:
   - nuget restore Bearded.Utilities.sln
-  - nuget install xunit.runners -Version 2.0.0 -OutputDirectory testrunner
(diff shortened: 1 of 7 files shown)


tomrijnbeek:
[Build Status](https://travis-ci.org/tomrijnbeek/utilities)

PR #36, opened 2015-04-10
improved random linq extensions

amulware:
including deferred shuffling and fixing shuffling distribution to be uniform

### DIFF
--- a/src/Linq/Extensions.cs
+++ b/src/Linq/Extensions.cs
@@ -133,12 +133,12 @@ namespace Bearded.Utilities.Linq
             if (random == null)
                 random = StaticRandom.Random;
             // optimisation for collections
-            var asList = source as ICollection<T>;
-            if (asList != null)
+            var asCollection = source as ICollection<T>;
+            if (asCollection != null)
             {
-                if (asList.Count == 0)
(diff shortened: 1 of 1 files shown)


src/Linq/Extensions.cs
```diff
@@ -248,6 +242,29 @@ public static IList<T> Shuffled<T>(this IEnumerable<T> source, Random random = n
             return list;
         }
 
+        /// <summary>
+        /// Returns a new shuffled list with the elements from the given sequence.
+        /// </summary>
+        /// <param name="source">The sequence to shuffle.</param>
+        /// <param name="random">An optional random object to be used. If none is given, StaticRandom is used instead.</param>
+        public static IEnumerable<T> ShuffledDeferred<T>(this IEnumerable<T> source, Random random = null)
```
tomrijnbeek: To me it is not entirely clear what this method does. I.e. what is the difference between this and just shuffling. For that reason, I think the comments have to be slightly more extensive.

amulware: I see your point. I sort of forgot about changing the comments. Will make sure to do so.

amulware:
Please review again. @tomrijnbeek

PR #47, opened 2015-09-13
Rectangular tilemap, and switching to .net 4.5.2

amulware:
This add basic rectangular tilemap, tile and direction types, and a bunch of extension methods to easily manage directions and bit fields of them (two different types).

There are also some tests to test the main functionality of the tilemap class.

Please review code, and let me know what you think.

We probably also want a bunch more tests for the other functionality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: csharp
+solution: Bearded.Utilities.sln
+install:
+  - nuget restore Bearded.Utilities.sln
+  - nuget install xunit.runner.console -Version 2.0.0 -OutputDirectory testrunner
+script:
+  - xbuild /p:Configuration=Release Bearded.Utilities.sln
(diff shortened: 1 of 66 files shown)


tomrijnbeek:
I would like to see the travis bot to be fixed before actually merging this if you don't mind. It doesn't like the switch to .NET 4.5.2.

amulware:
Certainly. Let me know if there is anything I can do to help.

tomrijnbeek:
The provided project file also does not compile on my own OSX device. I got it to work again by setting the target framework to 4.5 instead of 4.5.2.

```<TargetFrameworkVersion>v4.5</TargetFrameworkVersion>```

.NET version 4.5.2 is not provided by Mono, so it can't be used for cross platform development.

amulware:
Will travis work with 4.5? If so, I am fine making that change.

tomrijnbeek:
It should, yes.

amulware:
Will recreate targeting 4.5, also without all the erroneous commits.

Issue #45, opened 2015-08-29
IntervalOverlap is broken


tomrijnbeek:
Right now is exclusive, says is inclusive, durr.

PR #51, opened 2015-09-19
Math refactor

### LINKED ISSUE #41: GameMath -> Mathf
After working with Unity a bit more, I would suggest renaming the `GameMath` class to `Mathf`, or possibly even separating the basic math functions (`Cos`, `Sqrt`, etc.) into a `Mathf` class since they provide the same basic functionality as `Math` but on floats. This also saves a few characters in writing them in case that is important. Whether to include more "complicated" functions in `Mathf` or keep them in a separate `GameMath` class is open to discussion.

tomrijnbeek:
This pull request resolves #41, #45.

### DIFF
--- a/src/Bearded.Utilities.csproj
+++ b/src/Bearded.Utilities.csproj
@@ -125,7 +125,8 @@
     <Compile Include="Math\Direction2.cs" />
     <Compile Include="Math\Arc.cs" />
     <Compile Include="Math\Extensions.cs" />
-    <Compile Include="Math\GameMath.cs" />
+    <Compile Include="Math\Interval.cs" />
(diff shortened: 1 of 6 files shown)


amulware:
Any particular reason for replacing `float` with `Single`, in `Extensions.cs`?
For reference: https://github.com/tomrijnbeek/utilities/commit/cd8f6869403d0ba0c0f6df202144b63269159efb#diff-1d28989d80ddbe1a9c4380457840ccfaL288

tomrijnbeek:
That must have been Resharper changing things for me or something. Very odd. I can change it back if required.

tomrijnbeek:
Try this instead ;)

PR #59, opened 2016-03-26
Spacetime 2, closes #44

### LINKED ISSUE #44: Improve and finalize SpaceTime2 namespace
Prototype at https://github.com/amulware/ld33 can be used to determine a number of missing types and features.

Current work can be found at:
https://github.com/amulware/utilities/tree/spacetime2

amulware:
This includes the types:

- Unit, Speed, Acceleration (undirected)
- AngularVelocity, AngularAcceleration
- Position2, Difference2, Velocity2, Acceleration2
- Instant, Timespan
- Squared, IBackedBy

There is also some slight refactoring and improvement to existing math classes.
This code has been used extensively and seems to work great, but please review it nonetheless.
I will also get around to reading over it again soon, before we merge it.


Bearded.Utilities.sln
```diff
@@ -1,10 +1,17 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
+# Visual Studio 2013
+VisualStudioVersion = 12.0.31101.0
+MinimumVisualStudioVersion = 10.0.40219.1
```
tomrijnbeek: Why these changes?

Bearded.Utilities.sln
```diff
@@ -1,10 +1,17 @@
...
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Bearded.Utilities", "src\Bearded.Utilities.csproj", "{2F580B9D-6255-4427-8395-5EF236C9401C}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Bearded.Utilities.Tests", "tests\Bearded.Utilities.Tests.csproj", "{7EDB76CB-E9DC-4BCC-A610-12BE4976A701}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".nuget", ".nuget", "{EE2FBBE2-3CC6-4EE3-A4B1-FC026684B06C}"
+	ProjectSection(SolutionItems) = preProject
+		.nuget\packages.config = .nuget\packages.config
+	EndProjectSection
+EndProject
```
tomrijnbeek: Why these changes? I don't see an actual packages.config added to the repo in this PR.

src/Math/Angle.cs
```diff
@@ -44,7 +41,7 @@ public static Angle Between(Vector2 from, Vector2 to)
         {
             float perpDot = from.Y * to.X - from.X * to.Y;
 
-            return Angle.FromRadians(Mathf.Atan2(perpDot, Vector2.Dot(from, to)));
+            return FromRadians(Mathf.Atan2(perpDot, Vector2.Dot(from, to)));
```
tomrijnbeek: Did we decide to _not_ use full qualifiers for static methods now?

src/Math/Angle.cs
```diff
@@ -165,6 +162,7 @@ public Angle Abs()
         }
         /// <summary>
         /// Returns a new Angle with |value| == 1 radians and the same sign as this angle.
+        /// Returns 0 if the angle is zero.
```
tomrijnbeek: `Angle.Zero` I assume.

src/SpaceTime/2d/Acceleration2.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        #region construction
+
+        /// <summary>
+        /// Creates a new instance of the Acceleration2 type.
+        /// </summary>
+        public Acceleration2(Vector2 value)
+        {
+            this.value = value;
+        }
```
tomrijnbeek: Reason for missing whitespace?

src/SpaceTime/2d/Acceleration2.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        #region comparision
+
+        /// <summary>
+        /// Compares two acceleration vectors for equality.
+        /// </summary>
+        public static bool operator ==(Acceleration2 a0, Acceleration2 a1)
+        {
+            return a0.value == a1.value;
+        }
```
tomrijnbeek: I think it would be more consistent with how equals and comparison operators work if this operator would call the `Equals` method which has the implementation. The only reason to do it this way is to prevent an extra method call as overhead, but that could probably be mitigated by aggressive inlining.

src/SpaceTime/2d/Difference2.cs
```diff
@@ -0,0 +1,265 @@
...
+        {
+            return d0.value != d1.value;
+        }
+
+        #endregion
+
+        #endregion
+
+    }
+}
```
tomrijnbeek: Comments on `Acceleration2` apply here as well.

src/SpaceTime/2d/Position2.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            return p0.value != p1.value;
+        }
+
+        #endregion
+
+        #endregion
+
+    }
+}
```
tomrijnbeek: Comments on `Acceleration2` apply here as well.

src/SpaceTime/2d/Velocity2.cs
```diff
@@ -0,0 +1,285 @@
...
+        {
+            return v0.value != v1.value;
+        }
+
+        #endregion
+
+        #endregion
+
+    }
+}
```
tomrijnbeek: Comments on `Acceleration2` apply here as well.

src/SpaceTime/IBackedBy.cs
```diff
@@ -0,0 +1,17 @@
...
+    public interface IBackedBy<out T>
+        where T : struct
+    {
+        /// <summary>
+        /// Returns the numeric value of the type.
+        /// </summary>
+        T NumericValue { get; }
+    }
+
+}
```
tomrijnbeek: What would use cases for this interface be?

src/SpaceTime/Squared.cs
```diff
@@ -0,0 +1,169 @@
...
+        public static bool operator >=(Squared<T> s0, Squared<T> s1)
+        {
+            return s0.value >= s1.value;
+        }
+
+        #endregion
+
+        #endregion
+    }
+}
```
tomrijnbeek: Same comments form `Acceleration2` apply.

src/SpaceTime/angular/AngularAcceleration.cs
```diff
@@ -0,0 +1,232 @@
...
+        {
+            return v0.value >= v1.value;
+        }
+
+        #endregion
+
+        #endregion
+
+    }
+}
```
tomrijnbeek: Same comments form `Acceleration2` apply.

src/SpaceTime/angular/AngularVelocity.cs
```diff
@@ -0,0 +1,244 @@
...
+        {
+            return v0.value >= v1.value;
+        }
+
+        #endregion
+
+        #endregion
+
+    }
+}
```
tomrijnbeek: Same comments form `Acceleration2` apply.

src/SpaceTime/time/Instant.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            return i0.value >= i1.value;
+        }
+
+        #endregion
+
+        #endregion
+
+    }
+}
```
tomrijnbeek: Same comments form `Acceleration2` apply.

src/SpaceTime/time/TimeSpan.cs
```diff
@@ -0,0 +1,204 @@
...
+        {
+            return AngularVelocity.FromRadians(s.Radians / (float)t.value);
+        }
+
+        #endregion
+
+        #endregion
+
+    }
+}
```
tomrijnbeek: Same comments form `Acceleration2` apply.

src/SpaceTime/undirected/Acceleration.cs
```diff
@@ -0,0 +1,251 @@
...
+        {
+            return a0.value >= a1.value;
+        }
+
+        #endregion
+
+        #endregion
+
+    }
+}
```
tomrijnbeek: Same comments form `Acceleration2` apply.

src/SpaceTime/undirected/Speed.cs
```diff
@@ -0,0 +1,263 @@
...
+        {
+            return s0.value >= s1.value;
+        }
+
+        #endregion
+
+        #endregion
+
+    }
+}
```
tomrijnbeek: Same comments form `Acceleration2` apply.

src/SpaceTime/undirected/Unit.cs
```diff
@@ -0,0 +1,244 @@
...
+
+        /// <summary>
+        /// Returns a Unit type with value 0.
+        /// </summary>
+        public static Unit Zero { get { return new Unit(0); } }
+        /// <summary>
+        /// Returns a Unit type with value 1.
+        /// </summary>
+        public static Unit One { get { return new Unit(1); } }
+
```
tomrijnbeek: I think it may be a good idea to have a static "ofValue" method. That is: `Unit.ofValue(3)` or something similar. While there is `3.U()` to replace that, I think `Unit.ofValue(3)` is more descriptive and may be preferable in code being maintained by multiple people.

src/SpaceTime/undirected/Unit.cs
```diff
@@ -0,0 +1,244 @@
...
+        {
+            return u0.value >= u1.value;
+        }
+
+        #endregion
+
+        #endregion
+
+    }
+}
```
tomrijnbeek: Same comments form `Acceleration2` apply.

tomrijnbeek:
Please address comments.

src/SpaceTime/undirected/Unit.cs
amulware: You mean, you prefer `Unit.OfValue(3)` over `new Unit(3)`? (Also note that since these are value types, we cannot make the default constructor private, so instantiation with `new` will always be possible - if this is what you are concerned about.)

src/Math/Angle.cs
amulware: I do not remember, but did we agree to use full qualifiers before?

src/SpaceTime/2d/Acceleration2.cs
amulware: I'm not using empty lines between a lot of related methods (whether they be overloads, or a group of operators), basically at all, inside the same region (with a few exceptions, granted). Should I add empty lines everywhere?

src/SpaceTime/IBackedBy.cs
amulware: The idea is to allow for automatic generation of vertex attribute arrays from types that implement this interface, without the generator having to know all those types. This would allow users to define Position2 vectors, etc. in their vertex data, and have it be parsed correctly into a vec2 in GLSL for example. I can see the argument that maybe this should not be inside THIS library however... thoughts?

src/Math/Angle.cs
tomrijnbeek: I thought we did, especially based on how the math classes look. I guess I am fine with leaving them off, but then we should probable incrementally do that for the rest of the library as well.

src/SpaceTime/2d/Acceleration2.cs
tomrijnbeek: It appears to be the style of the rest of the library. Overloads I can see, but sometimes the grouping is a bit more subjective. I live separating the methods, I've never been one for the crammed look. It's up to you, vertical whitespace is always a bit of a weird thing, so I don't think it makes sense to put solid rules on that.

src/SpaceTime/IBackedBy.cs
tomrijnbeek: I understand. I think that's a very interesting thought, but I am not sure if an interface is the right tool. Actually, what this interface does is exposing implementation (the name very much implies to do so at least), and with implementing the interface you commit to keeping it that way. I suppose you could always rename the interface to something like `ReducesTo<T>` and make the `NumericValue` differently named as well that does not imply implementation.

Even then, the interface seems to be there to make a system in another library easier. The question is whether that is worth committing to an interface to.

src/SpaceTime/undirected/Unit.cs
tomrijnbeek: Ah, I am so not used to public constructors anymore, I forgot about that. I guess new Unit(3) also works, yes.

src/Math/Angle.cs
amulware: I think it makes sense to re-add the class names here. If we even want to refactor this, we should do it project wide in a separate pull request.

src/SpaceTime/2d/Acceleration2.cs
amulware: I'll see if I can separate them using a simple regex replace-all-in-files. Should not be too bad.

src/SpaceTime/IBackedBy.cs
amulware: I see your points. I will see if I can think of another solution. In the worst case, I could always list all known types in awgraphics. That means that I have to keep that list up to date, but in the end things should not change too often.

In fact, there is still a chance that I may change this entire system to use doubles instead of floats. That will require some performance tests first though. In either case, your point about exposing implementation is very interesting. However, in the end there has to be some sort of property to expose the underlying value to interact with other systems.

Again, I'll consider alternatives and make some proposed changes.

amulware:
I agree with all comments I did not reply to. I will address all points mentioned and create a new pull request soon.

PR #85, opened 2017-07-02
Clean up core namespace

amulware:
Mostly conversion to C#7 syntax and removal of silly XML documentation (I left things that actually explained non obvious things) and some minor refactoring.

Should not change behaviour in any way.

Squash as the bell strikes eleven.

### DIFF
--- a/src/Core/Do.cs
+++ b/src/Core/Do.cs
@@ -1,21 +1,16 @@
 ﻿namespace Bearded.Utilities
 {
-    /// <summary>
-    /// Collection of methods performing simple operations.
-    /// </summary>
     public static class Do
     {
-        /// <summary>
-        /// Swaps the values of the two variables.
-        /// </summary>
-        /// <typeparam name="T">The type of the two variables.</typeparam>
-        /// <param name="a">The first variable.</param>
-        /// <param name="b">The second variable.</param>
         public static void Swap<T>(ref T a, ref T b)
         {
             var t = a;
             a = b;
             b = t;
         }
+
+        public static Box<T> Box<T>(T value)
+            where T : struct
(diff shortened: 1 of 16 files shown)


src/Core/Do.cs
```diff
@@ -1,21 +1,16 @@
...
-        /// Swaps the values of the two variables.
-        /// </summary>
-        /// <typeparam name="T">The type of the two variables.</typeparam>
-        /// <param name="a">The first variable.</param>
-        /// <param name="b">The second variable.</param>
         public static void Swap<T>(ref T a, ref T b)
         {
             var t = a;
             a = b;
             b = t;
```
tomrijnbeek: Would (a, b) = (b, a) not work as replacement here?

src/Core/Do.cs
```diff
@@ -1,21 +1,14 @@
...
-        /// <param name="b">The second variable.</param>
         public static void Swap<T>(ref T a, ref T b)
         {
-            var t = a;
-            a = b;
-            b = t;
+            (a, b) = (b, a);
         }
+
+        public static Box<T> Box<T>(T value)
```
tomrijnbeek: This is an interface-breaking change. What is our policy on this?

src/Core/StaticRandom.cs
```diff
@@ -36,29 +35,17 @@ public static void SeedWith(int seed)
         /// <summary>
         /// Returns a random integer.
         /// </summary>
-        public static int Int()
-        {
-            return Random.Next();
-        }
+        public static int Int() => Random.Next();
```
tomrijnbeek: Let's set up some property tests for these!

In a separate PR, that is. Will make an issue.

PR #86, opened 2017-07-02
Add better thread-safe way of reading logger history

amulware:
Pretty straight forward I think.

Note the optimisation for lists, which will prevent resizing the list more than once.

Squash if you prefer not to allocate entire collections every frame.


amulware:
Ever heard of `.Clear()`?

PR #87, opened 2017-07-02
Add IIdable and IdDictionary

amulware:
Wanted to add a `Contains(T)` method to the dictionary, but was not sure if it should check only the id, or also get the value and call `Equals()`.

Squash if you have an id.


tomrijnbeek:
In case the behaviour is not obvious, leave it out, or give it a better name ;)

PR #91, opened 2017-07-07
Clean algos

tomrijnbeek:


src/Algorithms/HungarianAlgorithm.cs
```diff
@@ -195,28 +195,28 @@ private void reduce()
...
-            for (int s = 0; s < this.dim; s++)
+            for (var s = 0; s < dim; s++)
             {
-                for (int t = 0; t < this.dim; t++)
+                for (var t = 0; t < dim; t++)
                 {
-                    if (this.sourceMatches[s] == -1 && this.destMatches[t] == -1
-                        && this.costMatrix[s, t] - this.labelSources[s] - this.labelDests[t] == 0)
+                    if (sourceMatches[s] == -1 && destMatches[t] == -1
+                        && costMatrix[s, t] - labelSources[s] - labelDests[t] < float.Epsilon)
```
amulware: This looks like a change in behaviour to me? Even if the epsilon makes sense, what about negative values?

More importantly, what exactly is this doing that it checks `a - b - c` against zero?

tomrijnbeek: I am not even sure how this algorithm works. Reverted the change.

PR #92, opened 2017-07-07
Logger.Entry : IEquatable

amulware:
Clean commit is clean. 💅

### DIFF
--- a/src/Core/Logger.cs
+++ b/src/Core/Logger.cs
@@ -78,7 +78,7 @@ namespace Bearded.Utilities
             Trace = 5,
         }
         
-        public struct Entry
+        public struct Entry : IEquatable<Entry>
         {
             public string Text { get; }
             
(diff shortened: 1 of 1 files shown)


src/Core/Logger.cs
```diff
@@ -128,6 +128,28 @@ internal void WriteToConsole()
...
+            public bool Equals(Entry other)
+                => string.Equals(Text, other.Text) && Severity == other.Severity && Time.Equals(other.Time);
+
+            public override bool Equals(object obj) => obj is Entry && Equals((Entry)obj);
+
+            public override int GetHashCode()
+            {
+                unchecked
+                {
+                    var hashCode = Text != null ? Text.GetHashCode() : 0;
```
tomrijnbeek: Please replace with `Text?.GetHashCode() ?? null`

amulware: `?? 0` actually

tomrijnbeek: Right

PR #99, opened 2017-07-07
Make updates to Hungarian algorithm and add tests.

tomrijnbeek:


src/Algorithms/HungarianAlgorithm.cs
```diff
@@ -65,9 +67,8 @@ public static int[] Run(Vector3[] from, Vector3[] to)
         #endregion
 
         #region Implementation
-        #region Fields
         private readonly float[,] costMatrix;
-        private readonly int rows, cols, dim;
+        private readonly int dim;
```
amulware: Can't afford longer names?

src/Algorithms/HungarianAlgorithm.cs
```diff
@@ -76,19 +77,15 @@ public static int[] Run(Vector3[] from, Vector3[] to)
...
-        /// <summary>
-        /// Construct an instance of the algorithm.
-        /// </summary>
-        /// <param name="costMatrix">the cost matrix, where matrix[i][j] holds the cost of assigning worker i to job j, for all i, j.</param>
         private HungarianAlgorithm(float[,] costMatrix)
         {
+            if (costMatrix.GetLength(0) != costMatrix.GetLength(1))
+                throw new ArgumentException("Hungarian algorithm requires a square cost matrix.", nameof(costMatrix));
+            if (costMatrix.Cast<float>().Any(f => float.IsInfinity(f) || float.IsNaN(f)))
+                throw new ArgumentException("Only valid finite costs allowed.", nameof(costMatrix));
```
amulware: Be more explicit regarding !NaN requirement, instead of calling it 'valid'.

tests/Algorithms/HungarianAlgorithmTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using Xunit;
+
+namespace Bearded.Utilities.Tests.Algorithms
+{
+    public class HungarianAlgorithmTests
+    {
+        [Fact]
+        public void Run_EmptyCostMatrix_DoesntFail()
+        {
+            Assert.Empty(HungarianAlgorithm.Run(new float[0, 0]));
```
amulware: `DoesntFail` is not the same as checking `Empty`, is it? Should be renamed, changed, or split into two tests. Also please avoid contractions.

tests/Algorithms/HungarianAlgorithmTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        [Property]
+        public void Run_OneSource_AssignsTarget(float cost)
```
amulware: This test also tests two things. Please use custom generators (or parameterise the existing one if possible) and split into two tests.

tests/Algorithms/HungarianAlgorithmTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            if (float.IsInfinity(cost) || float.IsNaN(cost))
+                Assert.Throws<ArgumentException>(() => HungarianAlgorithm.Run(new[,] {{cost}}));
+            else
+                Assert.Equal(new[] { 0 }, HungarianAlgorithm.Run(new[,] { { cost } }));
+        }
+
+        [Fact]
+        public void Run_TwoSources_AssignsLowestCostTarget()
```
amulware: This should be a property test I think. Otherwise it does not really test what it claims.

src/Algorithms/HungarianAlgorithm.cs
tomrijnbeek: Done

PR #106, opened 2017-07-20
Add more Double math extensions

amulware:


src/Math/Extensions.cs
```diff
@@ -62,226 +47,173 @@ public static double Clamped(this double value, double min, double max)
...
-        /// <returns></returns>
-        public static int RoundedToInt(float f)
-        {
-            return (int)System.Math.Round(f);
-        }
+        public static int RoundedToInt(float f) => (int)System.Math.Round(f);
+
+        #endregion
+
+        #region Double math
```
tomrijnbeek: We have power functions like `Cubed` grouped by method, while trigonometry is grouped by float/double. Please make consistent.

tomrijnbeek: Not the actual power function though 🙄

src/Math/Extensions.cs
```diff
@@ -62,226 +47,173 @@ public static double Clamped(this double value, double min, double max)
...
         #region NaN sanity checks
         /// <summary>
+        /// Throws an exception if the specified double is NaN.
+        /// </summary>
+        public static void ThrowIfNaN(this double d,
+            string exceptionString = "Double is NaN while it is not allowed to.")
+        {
+            if (double.IsNaN(d))
+                throw new ArithmeticException(exceptionString);
+        }
```
tomrijnbeek: And back to grouping by method

amulware:
I'm not sure I see the issue with cleanup PRs. Improving code quality is a worth it on its own in my opinion.

amulware:
I organised everything by category (regions) and then method groups.

With the exception of the trig functions (but only those) which are still split into float/double regions, but have their own high-level category region, which I think makes it a bit clearer, since there are so many.

amulware:
Another semi important question: Do we want to remove comments from this file? Or at least from all the obvious things? `.Cos()` isn't really a big mystery.

tomrijnbeek:
Btw, I think I would still prefer the Trigonometry functions grouped by function instead of Float/Double, but it is not a blocker for you, so I leave it to your own discretion.

PR #107, opened 2017-07-23
Spacetime improvements

amulware:
Cleanup so far. Let me know if you want additions in here, or in a separate PR.

### DIFF
--- a/src/SpaceTime/2d/Acceleration2.cs
+++ b/src/SpaceTime/2d/Acceleration2.cs
@@ -13,25 +13,16 @@ namespace Bearded.Utilities.SpaceTime
 
         #region construction
 
-        /// <summary>
-        /// Creates a new instance of the Acceleration2 type.
-        /// </summary>
         public Acceleration2(Vector2 value)
         {
             this.value = value;
         }
-
-        /// <summary>
-        /// Creates a new instance of the Acceleration2 type.
-        /// </summary>
+        
         public Acceleration2(float x, float y)
             : this(new Vector2(x, y))
         {
         }
-
-        /// <summary>
-        /// Creates a new instance of the Acceleration2 type.
-        /// </summary>
+        
         public Acceleration2(Acceleration x, Acceleration y)
             : this(new Vector2(x.NumericValue, y.NumericValue))
         {
@@ -40,10 +31,7 @@ namespace Bearded.Utilities.SpaceTime
         /// <summary>
         /// Creates a new instance of the Acceleration2 type with a given direction and magnitude.
         /// </summary>
-        public static Acceleration2 In(Direction2 direction, Acceleration u)
-        {
-            return direction * u;
-        }
+        public static Acceleration2 In(Direction2 direction, Acceleration u) => direction * u;
 
         #endregion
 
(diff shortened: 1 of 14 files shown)


src/SpaceTime/2d/Acceleration2.cs
```diff
@@ -96,20 +84,14 @@ public static Acceleration2 In(Direction2 direction, Acceleration u)
         /// <param name="a0">The acceleration vector at t = 0.</param>
         /// <param name="a1">The acceleration vector at t = 1.</param>
         /// <param name="t">The interpolation scalar.</param>
-        public static Acceleration2 Lerp(Acceleration2 a0, Acceleration2 a1, float t)
-        {
-            return a0 + (a1 - a0) * t;
-        }
+        public static Acceleration2 Lerp(Acceleration2 a0, Acceleration2 a1, float t) => a0 + (a1 - a0) * t;
```
tomrijnbeek: nit: group static helper methods together

src/SpaceTime/angular/AngularVelocity.cs
```diff
@@ -48,47 +39,36 @@ public static AngularVelocity FromDegrees(float degrees)
...
         #region methods
 
         #region equality and hashcode
 
-        public bool Equals(AngularVelocity other)
-        {
-            return this.value == other.value;
-        }
+        // ReSharper disable once CompareOfFloatsByEqualityOperator
+        public bool Equals(AngularVelocity other) => value == other.value;
```
tomrijnbeek: Should we check `Math.abs(value - other.value) < float.Epsilon`? Is it possible to "propagate" the float comparison warning to the level using this method, since in a way it is still comparing two floats.

src/SpaceTime/2d/Acceleration2.cs
amulware: Where?

src/SpaceTime/angular/AngularVelocity.cs
amulware: That comparison basically does the same, for all values absolutely larger than ~1, and it will be much too insensitive for very small (absolutely) numbers.

More info: https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

TL;DR: Never use float quality after calculations. There are ways to fake it, but they aren't perfect. But that's besides the points. You should simply not use it for that.
What you can use it for us equality after assignment, which holds. This could be important if using the type as a key or similar.
In other words, I argue this implementation is correct.

amulware: Not to mention, it does the same as float.Equals(float) :P

src/SpaceTime/2d/Acceleration2.cs
tomrijnbeek: I would say at the bottom of the class in most cases, but it's not a hard rule.

amulware: This is the only static method here, and I rather have it grouped by functionality, instead of dangling around alone at the bottom.

amulware:
Can I haz review, answer to question in OP, or approval?

src/SpaceTime/2d/Acceleration2.cs
tomrijnbeek: Fair, though in other classes there are more static classes.

amulware: I went over all files, and it seems very consistent:
- constructing static functions follow constructors
- static properties (Zero, One) follow member properties
- lerp is grouped with relevant instance method
- operators are at the end of the file

I can see an argument for moving the static properties somewhere else, though I would then say the top of the file, not the bottom. If you want me to change that, let me know.

PR #109, opened 2017-07-29
Add StyleCop

amulware:
The first commit has all rules disabled (in the *.ruleset which is a MS/VS thing). The second enables one (spacing around keywords) and fixes them as example.

I think further rules should best be turned on (whether one or multiple at once depending on the amount of changes) and fixed in future PRs.


amulware:
We should also decide whether we want style warnings to be errors instead. That way compilation would fail entirely when rules are not followed.

tomrijnbeek:
I think ideally we want them to be warnings, but also surfaced through some form of CI; that way they can be ignored, but the reviewer will be informed as such and verify the decision that it can be ignored (sadly I don't think GitHub supports showing these things conveniently though). Though I feel that we should aim to never have warnings in our project at all if possible, so I wouldn't mind having the things as errors too much, but it makes them impossible to ignore if necessary, and to be honest, they are not really build errors.

amulware:
But, if a specific warning should be ignored in a specific case, we can add a suppression comment explaining the reason as well.

tomrijnbeek:
That is a practice we should use despite the severity of the message imo.

PR #110, opened 2017-08-04
:lipstick: Update Timespan operators to use doubles and clean class.

tomrijnbeek:


src/SpaceTime/time/TimeSpan.cs
```diff
@@ -10,34 +10,18 @@ public struct TimeSpan : IEquatable<TimeSpan>, IComparable<TimeSpan>
     {
         private readonly double value;
 
+        public double NumericValue => value;
```
amulware: It would be nice to have consistency in layout within the namespace, one way or the other.

PR #111, opened 2017-08-04
Refactor Hungarian algorithm

amulware:
I merely extracted the inner loops of the long and complicated `executePhase()` method into their own smaller methods. Please see if the names I chose make sense, I am not as familiar with all the terms involved.

### DIFF
--- a/src/Algorithms/HungarianAlgorithm.cs
+++ b/src/Algorithms/HungarianAlgorithm.cs
@@ -258,18 +258,7 @@ namespace Bearded.Utilities.Algorithms
         {
             while (true)
             {
-                int minSlackSource = -1, minSlackDest = -1;
-                var minSlackValue = float.PositiveInfinity;
-
-                for (int t = 0; t < dimension; t++)
-                {
-                    if (parentSourceByCommittedTarget[t] != -1) continue;
-                    if (!(minSlackValueBySource[t] < minSlackValue)) continue;
-
-                    minSlackValue = minSlackValueBySource[t];
-                    minSlackSource = minSlackTargetBySource[t];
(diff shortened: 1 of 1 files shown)


src/Algorithms/HungarianAlgorithm.cs
```diff
@@ -278,36 +267,64 @@ private void executePhase()
...
-                matchedSources[worker] = true;
-                for (var j = 0; j < dimension; j++)
-                {
-                    if (parentSourceByCommittedTarget[j] != -1) continue;
+                updateSlackValues(minSlackDest);
+            }
+        }
 
-                    var slack = costMatrix[worker, j] - sourceLabels[worker] - targetLabels[j];
+        private (int source, int destinatino, float value) findMinSlack()
```
tomrijnbeek: typo

src/Algorithms/HungarianAlgorithm.cs
```diff
@@ -278,36 +267,64 @@ private void executePhase()
...
+        {
+            var minSlackDest = -1;
+            var minSlackSource = -1;
+            var minSlackValue = float.PositiveInfinity;
 
-                    if (!(minSlackValueBySource[j] > slack)) continue;
+            for (var t = 0; t < dimension; t++)
+            {
+                if (parentSourceByCommittedTarget[t] != -1) continue;
+                if (!(minSlackValueBySource[t] < minSlackValue)) continue;
```
tomrijnbeek: why is this if inverted?

amulware: Ask the one who wrote it.

PR #118, opened 2017-08-20
Add ToString method to spacetime types and friends.

tomrijnbeek:
Feel free to suggest better formatting for to string methods.

### DIFF
--- a/src/Math/Angle.cs
+++ b/src/Math/Angle.cs
@@ -356,7 +356,7 @@ namespace Bearded.Utilities.Math
         public override string ToString() => ToString(null, CultureInfo.CurrentCulture);
 
         public string ToString(string format, IFormatProvider formatProvider)
-            => $"{radians.ToString(format, formatProvider)} rad";
+            => $"{Degrees.ToString(format, formatProvider)}°";
 
         #endregion
 
--- a/src/Math/Angle.cs
+++ b/src/Math/Angle.cs
@@ -93,28 +93,27 @@ namespace Bearded.Utilities.Math
         /// <summary>
         /// Gets the value of the angle in radians.
         /// </summary>
-        public float Radians { get { return this.radians; } }
+        public float Radians => this.radians;
 
         /// <summary>
         /// Gets the value of the angle in degrees.
         /// </summary>
-        public float Degrees { get { return Mathf.RadiansToDegrees(this.radians); } }
+        public float Degrees => Mathf.RadiansToDegrees(this.radians);
 
         /// <summary>
         /// Gets a 2x2 rotation matrix that rotates vectors by this angle.
         /// </summary>
-        public Matrix2 Transformation { get { return Matrix2.CreateRotation(this.radians); } }
+        public Matrix2 Transformation => Matrix2.CreateRotation(this.radians);
 
         /// <summary>
         /// Gets the magnitude (absolute value) of the angle in radians.
         /// </summary>
-        public float MagnitudeInRadians { get { return System.Math.Abs(this.radians); } }
+        public float MagnitudeInRadians => System.Math.Abs(this.radians);
 
         /// <summary>
         /// Gets the magnitude (absolute value) of the angle in degrees.
         /// </summary>
-        public float MagnitudeInDegrees { get { return System.Math.Abs(this.Degrees); } }
-
+        public float MagnitudeInDegrees => System.Math.Abs(this.Degrees);
 
         #endregion
         
@@ -348,6 +347,12 @@ namespace Bearded.Utilities.Math
         {
             return x.radians >= y.radians;
         }
+
+        #endregion
+
+        #region String
+
+        public override string ToString() => $"{radians} rad";
 
         #endregion
 
--- a/src/Math/Angle.cs
+++ b/src/Math/Angle.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Globalization;
 using OpenTK;
 
 namespace Bearded.Utilities.Math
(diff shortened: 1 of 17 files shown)


tomrijnbeek:
Codeacy issues were caused by some existing floating-point comparison problems being unignored due to changes around them. I reignored them and rerun should show green.
Travis... I have no idea what is going on. Tests are failing with a SIGSEV, but they run fine locally. They still failed after triggering the tests again. I'll see what happens when trying to run them on develop.

tomrijnbeek:
I fixed travis by changing the used mono version to alpha. Not idea maybe, but apparently necessary. I am happy to drop travis entirely, since the only thing it adds over Appveyor is the Linux environment, but it has also been the most brittle CI we have.

amulware:
Does appveyor run the tests?

src/Math/Angle.cs
```diff
@@ -351,6 +350,12 @@ public override int GetHashCode()
 
         #endregion
 
+        #region String
+
+        public override string ToString() => $"{radians} rad";
```
amulware: I think these should return degrees for better human readability. Alternatively, consider extracting the factor of pi?

src/Math/Direction2.cs
```diff
@@ -245,6 +245,12 @@ public override int GetHashCode()
 
         #endregion
 
+        #region tostring
+
+        public override string ToString() => $"{Radians} rad";
```
amulware: Ditto. Also an option: Haven an overload with a parameter (create an enum) to either print degrees or radians or radians with the factor of pi extracted.

src/Math/Geometry/PolarPosition.cs
```diff
@@ -104,14 +93,17 @@ public override bool Equals(object obj)
...
                 return 0;
 
             unchecked
             {
-                return (this.r.GetHashCode() * 397) ^ this.angle.GetHashCode();
+                return (R.GetHashCode() * 397) ^ Angle.GetHashCode();
             }
         }
+
+        public override string ToString() => $"{R} ∠{Angle}";
```
amulware: What's the reasoning with the space?

src/SpaceTime/Squared.cs
```diff
@@ -78,6 +78,12 @@ public static Squared<T> FromValue(float value)
         public int CompareTo(Squared<T> other) => value.CompareTo(other.value);
 
         #endregion
+        
+        #region tostring
+
+        public override string ToString() => $"|{value}|";
```
amulware: This no longer has any idea of its actual unit then?

src/SpaceTime/angular/AngularAcceleration.cs
```diff
@@ -66,6 +66,12 @@ private AngularAcceleration(float value)
 
         #endregion
 
+        #region tostring
+
+        public override string ToString() => $"{value} rad/t²";
```
amulware: Degrees vs radians?

src/SpaceTime/angular/AngularVelocity.cs
```diff
@@ -65,6 +65,12 @@ private AngularVelocity(float value)
         public override int GetHashCode() => value.GetHashCode();
 
         #endregion
+        
+        #region tostring
+
+        public override string ToString() => $"{value} rad/t";
```
amulware: Ditto.

src/Math/Direction2.cs
tomrijnbeek: I think I'd rather just have sensible defaults, especially since ToString() is what will be used to show debug/logging information, and custom overloads won't really be picked up by anything unless you call them specifically.

src/Math/Geometry/PolarPosition.cs
tomrijnbeek: Readability

src/SpaceTime/Squared.cs
tomrijnbeek: I couldn't find a way to get to the `ToString` method from the `T` type. Feel free to let me know if you have any ideas.

amulware: Only gross ones right now...

src/Math/Direction2.cs
amulware: So you'd rather have people do `direction.Radians.ToString()` if they wanna be specific? I guess that is no more code than otherwise, though you then lose the unit, but you can just do $"{direction.Radians} rad" yourself. I feel that Degrees might be a better default than Radians though, since most people will probably have an easier time with that? Or at least the factor of PI thing? Radians are just.. hard to understand. 6.28...

tomrijnbeek:
Let's go with degrees and add a degree symbol. That should be good enough
for debugging purposes. If you want a fancy display somewhere, you probably
want to do it manually anyway

src/SpaceTime/2d/Position2.cs
```diff
@@ -80,6 +80,13 @@ public Position2(Unit x, Unit y)
 
         #endregion
 
+        #region tostring
+
+        public string ToString(string format, IFormatProvider formatProvider)
```
amulware: Missing the other method.

src/SpaceTime/2d/Velocity2.cs
```diff
@@ -126,6 +126,15 @@ public Velocity2(Speed x, Speed y)
 
         #endregion
 
+        #region tostring
+
+        public override string ToString() => $"{value} u/t";
```
amulware: This should call the other method, right?

Issue #108, opened 2017-07-29
Remove Core's dependency on Linq


tomrijnbeek:
The `Core` and `Linq` namespaces form a dependency cycle. In my opinion `Core` should not depend on anything and stand on its own, so we should move the code that depends on `Linq` somewhere else.

![image](https://user-images.githubusercontent.com/828553/28740191-ea170cd2-73f9-11e7-9836-d5a684c3b5ff.png)

In particular, this is the `IdManager` using the `ValueOrDefault` extension.

![image](https://user-images.githubusercontent.com/828553/28740218-33c3d77a-73fa-11e7-9edb-b345a5f4ee7a.png)

amulware:
Sure. How come Core relies on Math? That seems strange.

tomrijnbeek:
It seems like Logger relies on Math for some reason. A quick check shows we use the `.Clamped` extension. We can probably just replace that with a `System.Math.Clamp`.

amulware:
System.Math.Clamp is a thing? Huh.

tomrijnbeek:
Oh no, it's not :(

PR #120, opened 2017-10-15
Remove Core dependency on Linq. See #108

tomrijnbeek:


src/Core/IdManager.cs
```diff
@@ -11,7 +10,15 @@ public sealed class IdManager
...
-            var id = lastIds.ValueOrDefault(type) + 1;
+            int id;
+            if (lastIds.TryGetValue(type, out var lastId))
+            {
+                id = lastId + 1;
+            }
+            else
+            {
+                id = 1;
+            }
```
amulware: I think I would prefer `?:` syntax.

Issue #88, opened 2017-07-03
Add property tests for math functions


tomrijnbeek:
We talked about adding property tests, and the math functions seem to be prime candidates for that. I think it would be great if we could sanity check the Mathf and Random functions at least.

amulware:
By implementing the two lines of each method again in the test and comparing the results? :/

Not even sure how to test Random, except to make sure that only valid numbers are returned, but I don't wanna write statistical tests to check if the result is actually random.

tomrijnbeek:
I was indeed thinking about just sanity checking that the output lies within the supplied range.

PR #124, opened 2017-10-23
Add random extension tests

amulware:
See #88.

Currently, several tests fail, mostly because we have not really defined the behaviour of our NextLong and NextDouble methods.

I propose we use the same behaviour as System.Random.Next (returning integers):

- NextX(max) throws if max is negative, returns 0 if max is 0, and otherwise samples from [0, max[, 
- NextX(min, max) throws if max is less than min, returns min if min is max, and otherwise samples from [min, max[, 

I don't really _like_ this behaviour, especially for the max==0/min==max, but I suppose this would be the most consistent with existing behaviours?

I'm also not sure how to test the normally distributed methods. Should I just call them all to make sure they don't crash? Maybe check that they don't return NaN/+-infinity?

### DIFF
--- a/tests/Core/RandomExtensionTests.cs
+++ b/tests/Core/RandomExtensionTests.cs
@@ -1,0 +1,147 @@
+﻿using System;
+using FsCheck.Xunit;
+using Xunit;
+
+namespace Bearded.Utilities.Tests.Core
+{
+    public class RandomExtensionTests
+    {
+        [Property]
+        public void NextLong(int seed)
+        {
+            var random = new Random(seed);
+
+            random.NextLong();
+        }
+
+        [Property]
+        public void NextLongWithMax(int seed, long max)
+        {
+            var random = new Random(seed);
+
+            var number = random.NextLong(max);
+            
+            Assert.InRange(number, 0, max - 1);
+        }
+
+        [Property]
+        public void NextLongWithMinAndMax(int seed, long min, long max)
+        {
+            var random = new Random(seed);
+
+            var number = random.NextLong(min, max);
+
+            Assert.InRange(number, min, max - 1);
+        }
+
+        [Property]
+        public void NextDoubleWithMax(int seed, double max)
+        {
+            var random = new Random(seed);
+
+            var number = random.NextDouble(max);
+
+            Assert.InRange(number, 0, max);
+            Assert.NotEqual(max, number);
+        }
+
+        [Property]
+        public void NextDoubleWithMinAndMax(int seed, double min, double max)
+        {
+            var random = new Random(seed);
+
+            var number = random.NextDouble(min, max);
+
+            Assert.InRange(number, 0, max);
+            Assert.NotEqual(max, number);
+        }
+
+        [Property]
+        public void NextFloat(int seed)
+        {
+            var random = new Random(seed);
+
+            random.NextFloat();
+        }
+
+        [Property]
+        public void NextFloatWithMax(int seed, float max)
+        {
+            var random = new Random(seed);
+
+            var number = random.NextFloat(max);
+
+            Assert.InRange(number, 0, max);
+            Assert.NotEqual(max, number);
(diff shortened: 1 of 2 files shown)


tests/Core/RandomExtensionTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            var random = new Random(seed);
+
+            var number = random.NextLong(max);
+            
+            Assert.InRange(number, 0, max - 1);
+        }
+
+        [Property]
+        public void NextLongWithMinAndMax(int seed, long min, long max)
```
tomrijnbeek: What happens if max < min? I find it a bit confusing to see a property test testing min and max without any restrictions on the generated min/max. Should we create a property generator for (min, max) interval pairs?

tests/Core/RandomExtensionTests.cs
```diff
@@ -0,0 +1,147 @@
...
+            Assert.InRange(number, 0, max);
+            Assert.NotEqual(max, number);
+        }
+
+        [Property]
+        public void NextFloat(int seed)
+        {
+            var random = new Random(seed);
+
+            random.NextFloat();
```
tomrijnbeek: Same for NextLong above I suppose, but I find it a bit odd to have a test without a single assertion. Should we maybe at least check if the result is a valid float? Assert.IsNotNaN or something of the like?

amulware:
I can see where you are coming from, but before we fix the tests, I would like to properly define intended behaviour (see above).

amulware:
@tomrijnbeek any comment on the above?

tomrijnbeek:
I think for integral number types, we should definitely mimic the behaviour of the existing `Next` methods. While extension methods are different, they are sort of presented in a similar way as the normal public methods, and I believe it would lead to confusion if their behaviour was different in a significant way.

If we want to have different behaviour, I believe it should be presented in a different way to make the behaviour difference clear by the different ways you are calling the method.

I do think the behaviour makes sense. The documentation specifies _A 32-bit signed integer greater than or equal to minValue and less than maxValue; that is, the range of return values includes minValue but not maxValue. If minValue equals maxValue, minValue is returned._

For floating point types, I would argue the same thing: we should follow the existing behaviour wherever possible.

amulware:
The question about floating point number is: There is only one official method which returns a value in `[0, 1[`

So we have no precedence for `NextDouble(max)` and `NextDouble(min, max)` where `max <= min/0`.

The calculations internally are `NextDouble() * max` and `NextDouble() * (max - min) + min`.

I can't decide if these should throw to be in line with `NextInt` or should just return the calculated value, it's not like it's badly defined...

Another option: Maybe don't call them `min` and `max` but `inclusiveBound` and `exclusiveBound`? Then we don't have to worry about throwing.

tomrijnbeek:
I don't think `inclusiveBound` and `exclusiveBound` are clear enough to be used as parameter names. If you just look at those parameter names, you'd have to read the documentation to understand what the function does.

I do understand your comment about how it is well defined though. What about `NextDoubleOnInterval(double startInclusive, double endExclusive)`? This way the function makes it explicitly clear that we are looking at `[startInclusive, endExclusive[`, and if the caller site decides that `[10, 4[` is a valid interval, it'll just work.

In general, I'm okay with not making it throw, as long as we can make the well-defined behaviour here clear from the method interface, ideally without extensive documentation. Does that sound reasonable?

amulware:
Alright, I'll go with that. Not throwing, and fewer branches is generally better for most of our use cases anyways.

I don't see why my proposed names are not clear though. The method is sampling an interval (that's general knowledge), which has two bounds (is this not obvious enough), one of which is included, the other is excluded (your names are the same here).

I don't see what 'start' and 'end' add, except that they imply order which really doesn't exist as far as I can tell.

amulware:
Closing this. A replacement is WIP.

PR #131, opened 2018-06-24
:racehorse: Create logger constructor that only instantiates specified logger writers

### LINKED ISSUE #125: Make Logger severity interface nullable.
If you currently do `Logger.Debug.Log($"this is a log {expensiveOperation()}")`, it will evaluate the string even in production code. We should make the different severity levels nullable, and change the public interface to `Logger.Debug?.Log($"...")`. This will stop the strings from being evaluated if `Debug` is set to `null` (i.e. in production).

tomrijnbeek:
This can be used in combination with changing the logging calls to `logger.Debug?.Log()`. If you compile a production binary, you can not instantiate the writers for debug and trace (or even warning). In the case of string interpolation, this will cause a speed up of the code.

For example, if in your code you have

```
logger.Debug.Log($"Mouse coordinates at {mouse.Coordinates}");
```

The string will be interpolated even if you disable the debugging lines in your log & console. However, by changing the line to:

```
logger.Debug?.Log($"Mouse coordinates at {mouse.Coordinates}");
```

The string will only be interpolated if the Debug writer is actually instantiated.

The change is backwards compatible, since all logger writers will be instantiated in the default constructor, meaning null-checks are unnecessary. That being said, I suggest changing the writers to be explicitly nullable whenever C#8 comes around.

Resolves #125

### DIFF
--- a/src/IO/Logger.cs
+++ b/src/IO/Logger.cs
@@ -135,7 +135,7 @@ namespace Bearded.Utilities.IO
             public bool Equals(Entry other)
                 => string.Equals(Text, other.Text) && Severity == other.Severity && Time.Equals(other.Time);
 
-            public override bool Equals(object obj) => obj is Entry && Equals((Entry)obj);
+            public override bool Equals(object obj) => obj is Entry entry && Equals(entry);
 
             public override int GetHashCode()
             {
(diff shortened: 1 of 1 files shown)


tomrijnbeek:
Sure. The writers are now explicitly nullable. Happy to add an annotation, but since we don't do that anywhere so far, I guess we're waiting for C#8 😄

PR #132, opened 2018-10-13
:sparkles: Add DirectedAcyclicGraph class

tomrijnbeek:
Adds a directed acyclic graph class. This seems to be the easiest way to represent a partial ordering, which is something I need for an algorithm I want to add to this library.


tomrijnbeek:
Created a new PR that I think implements this better.

PR #135, opened 2018-10-13
Use text config for appveyor

tomrijnbeek:


tomrijnbeek:
This is literally downloading the text file from their website.

Reason for implementing is that we may want to make changes based on a PR, and it would be awkward if the config change worked for only that branch or everything but that branch.

PR #136, opened 2018-10-29
Implement directed graph data structures

tomrijnbeek:
This is a replacement for #132 favouring immutable structures and fluent method chains.


src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+        public static DirectedGraphBuilder<T> FromExistingGraph(IDirectedGraph<T> graph)
+        {
+            var builder = new DirectedGraphBuilder<T>();
+
+            foreach (var element in graph.Elements)
+            {
+                builder.elements.Add(element);
+
+                var successors = new HashSet<T>(graph.GetDirectSuccessorsOf(element));
+                var predecessors = new HashSet<T>(graph.GetDirectPredecessorsOf(element));
```
amulware: So much copying of data. 😱

src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+            return new DirectedGraphBuilder<T>().CreateAcyclicGraphUnsafe();
+        }
+
+        private DirectedGraphBuilder() { }
+
+        public DirectedGraphBuilder<T> AddElement(T element)
+        {
+            if (element == null) throw new ArgumentNullException(nameof(element));
+
+            if (!elements.Add(element))
```
amulware: Is there utility to swallowing double adds silently?

src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            if (sources.Contains(to))
+            {
+                sources.Remove(to);
+            }
+
+            return this;
+        }
+
+        public DirectedGraphBuilder<T> RemoveElement(T element)
```
amulware: A builder that allows removal? Why?

src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            if (element == null) throw new ArgumentNullException(nameof(element));
+
+            if (!elements.Remove(element))
```
amulware: Another possible no-op?

src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        public DirectedGraphBuilder<T> AddArrow(T from, T to)
+        {
+            if (from == null) throw new ArgumentNullException(nameof(from));
+            if (to == null) throw new ArgumentNullException(nameof(to));
+            if (!elements.Contains(from)) throw new ArgumentException("Element not in graph.", nameof(from));
+            if (!elements.Contains(to)) throw new ArgumentException("Element not in graph.", nameof(to));
+            if (from.Equals(to))
+            {
+                throw new InvalidOperationException("Cannot make self-edges in directed acyclic graph.");
```
amulware: The class name doesn't say that this builds acyclic graphs. 

The method also doesn't prevent other cycles, does it?

src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            if (element == null) throw new ArgumentNullException(nameof(element));
+
+            if (!elements.Remove(element))
+            {
+                return this;
+            }
+            sources.Remove(element);
+
+            foreach (var to in directSuccessors[element])
```
amulware: I could probably understand this loop with more thinking or more sleep prior to reviewing, but I suggest it might want to have a name.

src/Graphs/DirectedGraphBuilder.cs
amulware: After reading more below, it looks like this graph can build both, cyclic and acyclic graphs - should this check be here then?

src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            return this;
+        }
+
+        public DirectedGraphBuilder<T> RemoveArrow(T from, T to)
+        {
+            if (from == null) throw new ArgumentNullException(nameof(from));
+            if (to == null) throw new ArgumentNullException(nameof(to));
+            if (!elements.Contains(from)) throw new ArgumentException("Element not in graph.", nameof(from));
+            if (!elements.Contains(to)) throw new ArgumentException("Element not in graph.", nameof(to));
```
amulware: Do we need these two checks? Can't we just see if the edge is known?

I'm a bit confused why it will fail if the elements aren't known, but if the edge isn't known it's fine.

src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            if (isGraphCyclic())
+            {
+                throw new InvalidOperationException("Cannot create an directed acyclic graph with cycles.");
+            }
+
+            return CreateAcyclicGraphUnsafe();
+        }
+
+        private bool isGraphCyclic()
```
amulware: Worth exposing publicly? /shrug

src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            var visited = new HashSet<T>();
+            var currentStack = new HashSet<T>();
+
+            return elements.Where(checkGraphCycleUsingDepthFirst).Any();
```
amulware: Isn't this more easy to understand as a regular loop(, or maybe recursive call)?

src/Graphs/IDirectedAcyclicGraph.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Bearded.Utilities.Graphs
+{
+    public interface IDirectedAcyclicGraph<T> : IDirectedGraph<T> where T : IEquatable<T>
+    {
+        IDirectedAcyclicGraph<T> GetTransitiveReduction();
```
amulware: Should reasonably complicated logic like this be part of the data structure itself? Is this a property of the data or something the data can do?

tomrijnbeek: Removed and made the transitive reducer public. I guess we can still make it an extension method and have the same effect though...

src/Graphs/DirectedGraphBuilder.cs
tomrijnbeek: I don't think so, but feel free to convince me otherwise.

src/Graphs/DirectedGraphBuilder.cs
tomrijnbeek: Definitely not a regular loop. Implementing depth-first using a regular loop will only become messy. The recursive call is actually what happens in the method below. I could pull it out, but I don't think the logic would change whatsoever. We would just pass in `visited` and `currentStack` as parameters instead. Feel free to suggest an alternative solution though.

src/Graphs/IDirectedAcyclicGraph.cs
amulware: You say removed, but this is still here. Not sure I follow. 🤔

src/Graphs/DirectedGraphBuilder.cs
amulware: After stepping through it, I see why the complexity of the algorithm is necessary. I tried writing a version without `currentStack`, but it is needed to not interpret converging paths as cycles while also being optimal (O(n+m)).

An O(n log n + m) algorithm can be a lot prettier (sort elements by number of predecessors in a priority queue, repeatedly dequeue the top if it has no predecessors, until the priority queue is empty, or the top element has a predecessor - i.e. topologically sorting the graph), but it _is_ less efficient. :(

amulware: I still think this can be made more readable. I'll write several proposals.

src/Graphs/AdjacencyListDirectedGraph.cs
```diff
@@ -0,0 +1,29 @@
...
+            ImmutableList<T> elements,
+            ImmutableDictionary<T, ImmutableList<T>> directSuccessors,
+            ImmutableDictionary<T, ImmutableList<T>> directPredecessors)
+        {
+            this.elements = elements;
+            this.directSuccessors = directSuccessors;
+            this.directPredecessors = directPredecessors;
+        }
+
+        public IEnumerable<T> GetDirectSuccessorsOf(T element) => directSuccessors[element];
```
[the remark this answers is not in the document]
tomrijnbeek: Yes, though I agree this should be explicit. Changed.

src/Graphs/IDirectedAcyclicGraph.cs
tomrijnbeek: Fixed

src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,158 @@
...
+            if (!elements.Add(element)) throw new ArgumentException("Element already in graph.", nameof(element));
+
+            sources.Add(element);
+            directSuccessors.Add(element, new HashSet<T>());
+            directPredecessors.Add(element, new HashSet<T>());
+
+            return this;
+        }
+
+        public DirectedGraphBuilder<T> AddArrow(T from, T to)
```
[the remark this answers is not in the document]
tomrijnbeek: Yes, since edge explicitly refers to an undirected connection, and arc is ambiguous. Suggestions welcome, otherwise... resolved.

src/Graphs/DirectedGraphBuilder.cs
```diff
@@ -0,0 +1,158 @@
...
+                currentStack.Remove(element);
+
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Creates an immutable directed acyclic graph instance without verifying that there are no cycles.
+        /// </summary>
+        public IDirectedAcyclicGraph<T> CreateAcyclicGraphUnsafe()
```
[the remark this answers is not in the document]
tomrijnbeek: If you explicitly construct your own DAG, and you trust your own logic, I don't want to enforce having to do a full O(n) check on validating the graph.

src/Graphs/DirectedGraphBuilder.cs
tomrijnbeek: Took part of the changes

src/Graphs/DirectedAcyclicGraphTransitiveReducer.cs
```diff
@@ -0,0 +1,72 @@
...
+                }
+            }
+
+            return graphBuilder.CreateAcyclicGraphUnsafe();
+        }
+
+        private static IEnumerable<T> getOnlyDirectChildren(IDirectedGraph<T> graph, T element)
+        {
+            var children = graph.GetDirectSuccessorsOf(element) as IList<T>
+                ?? graph.GetDirectSuccessorsOf(element).ToList();
```
amulware: I'd prefer caching the result of this method I think. Though we do much worse than this in the recursive function below I suppose...

src/Graphs/DirectedAcyclicGraphTransitiveReducer.cs
```diff
@@ -0,0 +1,72 @@
...
+                foreach (var neighbor in getOnlyDirectChildren(graph, element))
+                {
+                    graphBuilder.AddArrow(element, neighbor);
+                }
+            }
+
+            return graphBuilder.CreateAcyclicGraphUnsafe();
+        }
+
+        private static IEnumerable<T> getOnlyDirectChildren(IDirectedGraph<T> graph, T element)
```
amulware: I think these methods should only take acyclic graphs for correctness.

src/Graphs/DirectedAcyclicGraphTransitiveReducer.cs
```diff
@@ -0,0 +1,72 @@
...
+                {
+                    yield return descendant;
+                }
+            }
+        }
+    }
+
+    public static class DirectedAcyclicGraphTransitiveReducer
+    {
+        public static IDirectedAcyclicGraph<T> ReduceGraph<T>(IDirectedAcyclicGraph<T> graph)
```
amulware: Could be an extension method?

amulware: Though then it's losing the 'transitive' semantic.. so many not.

src/Graphs/DirectedAcyclicGraphTransitiveReducer.cs
```diff
@@ -0,0 +1,72 @@
...
+            var graphBuilder = DirectedGraphBuilder<T>.NewBuilder();
+
+            foreach (var element in elements)
+            {
+                graphBuilder.AddElement(element);
+            }
+
+            foreach (var element in elements)
+            {
+                foreach (var neighbor in getOnlyDirectChildren(graph, element))
```
amulware: I'd expect the iterator to be called `child`, given the name of the method.

src/Graphs/DirectedAcyclicGraphTransitiveReducer.cs
amulware: Wouldn't `getDirectOnlyChildren` be a better name? Or maybe `getDescendantsThatCanOnlyBeReachedDirectly` or something?

src/Graphs/AdjacencyListDirectedGraph.cs
```diff
@@ -0,0 +1,35 @@
...
+    class AdjacencyListDirectedGraph<T> : IDirectedGraph<T> where T : IEquatable<T>
+    {
+        private readonly ImmutableList<T> elements;
+        private readonly ImmutableDictionary<T, ImmutableList<T>> directSuccessors;
+        private readonly ImmutableDictionary<T, ImmutableList<T>> directPredecessors;
+
+        public IEnumerable<T> Elements => elements;
+        public int Count => elements.Count;
+
+        public AdjacencyListDirectedGraph(
```
amulware: Consider making this internal so nobody can sub-class this outside the binary?

src/Graphs/AdjacencyListDirectedGraph.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            this.elements = elements;
+            this.directSuccessors = directSuccessors;
+            this.directPredecessors = directPredecessors;
+        }
+
+        public IEnumerable<T> GetDirectSuccessorsOf(T element) =>
+            directSuccessors.ContainsKey(element)
+                ? directSuccessors[element]
+                : throw new ArgumentOutOfRangeException(nameof(element), "Element not found in graph.");
```
amulware: I really don't like double access. Could we TryGetValue instead?
```cs
directSuccessors.TryGetValue(element, out var successors)
    ? successors
    : throw new ArgumentOutOfRangeException(nameof(element), "Element not found in graph.");
```

(Same below.)

src/Graphs/DirectedAcyclicGraphTransitiveReducer.cs
tomrijnbeek: That's my thought

tests/Graphs/DirectedGraphBuilderTests.cs
```diff
@@ -0,0 +1,270 @@
...
+
+            graph.Count.Should().Be(1);
+        }
+
+        [Fact]
+        public void AddElement_ShouldThrowOnNull()
+        {
+            var builder = DirectedGraphBuilder<string>.NewBuilder();
+
+            builder.Invoking(b => b.AddElement(null)).Should().Throw<ArgumentNullException>();
```
amulware: I prefer something like:
```cs
Action callingAddElementWithNull = () => builder.AddElement(null);

callingAddElementWithNull.Should().Throw();
```

Consider it, but I won't block the PR for this.

src/Graphs/AdjacencyListDirectedGraph.cs
tomrijnbeek: There classes themselves are already internal, but done anyway in case that changes in the future.

src/Graphs/DirectedAcyclicGraphTransitiveReducer.cs
tomrijnbeek: The latter feels too long. This method "only gets the direct children". `getOnlyDirectChildren` sounds like the best translation of that in a method name. I wouldn't know what "direct only" would mean

src/Graphs/DirectedAcyclicGraphTransitiveReducer.cs
tomrijnbeek: This is caused by a previous naming. Will update

src/Graphs/DirectedAcyclicGraphTransitiveReducer.cs
amulware: Maybe it is clear enough if you know what the algorithm is trying to do.

tomrijnbeek: Changed

tests/Graphs/DirectedGraphBuilderTests.cs
```diff
@@ -91,15 +92,17 @@ public void AddElement_ShouldThrowOnNull()
         {
             var builder = DirectedGraphBuilder<string>.NewBuilder();
 
-            builder.Invoking(b => b.AddElement(null)).Should().Throw<ArgumentNullException>();
+            Action addNullElement = () => builder.AddElement(null);
+            addNullElement.Should().Throw<ArgumentNullException>();
```
amulware: Empty lines between these two please. :) And similar for all other tests below.

PR #137, opened 2018-12-14
Make DeletableObjectList safer

### LINKED ISSUE #61: DeletableObjectList ForceCompact() and TrimExcess() could break enumeration
Specifically, since these methods do not check if the list is currently being enumerated, and active enumerators may skip elements, which could lead to unpredictable and undesirable behaviour.

The two methods should check if there are active enumerators and throw if so.

amulware:
This makes several operations in the DeletableObjectList class that could lead to an invalid state throw instead.

There is also some minor clean up.

Closes #61


src/Collections/DeletableObjectList.cs
```diff
@@ -166,7 +171,7 @@ public void ForceCompact()
         /// </summary>
         public void TrimExcess()
```
tomrijnbeek: Style nit: I prefer methods calling down. Can we move `TrimExcess` above `ForceCompact`?

tomrijnbeek:
Looks like the style nit is the only open comment.

Please update the base branch to master and update it. I'll approve once that's done :)

PR #144, opened 2019-01-18
Remove unnecessary method

amulware:
Might as well just do what this method does wherever this is used.


amulware:
Updated the one usage of the method. If you don't like it, close it.

PR #149, opened 2019-01-20
Add types for gravity calculations

### LINKED ISSUE #143: Support gravity calculations
This might be complicated (given the unit of the gravitational constant) but worth doing if we can do it right.

Might also require types for mass and force.

amulware:
Closes #143

The main motivation here is that doing calculations with gravity are not entirely trivial, and easily messed up. Without types for them, there is no easy way to make sure that the units work out correctly, and that the calculations are done correctly.

Given the complex unit of the gravitational constant (m^3 kg^-1 s^-2), I decided it would be best to hide all of this complexity inside a single type (aptly named `GravitationalConstant`). That way the only other type that was needed was `Mass`, all the forces and complex units of intermediate values are hidden inside a single calculation in `GravitationalConstant.AccelerationAtDistance(Mass, Squared<Unit>)`.

The `Mass` type itself is a simple ordered linear space and the `GravitationalConstant` class represents a single constant allowing for several gravity calculations, all of which return directed or undirected accelerations (the type for that already exists).

Example use in an object attracted by another would be:

```
// somewhere in game setup
G = new GravitationalConstant(0.37f);

// update object
var acceleration = G.AccelerationTowards(other.Mass, other.Position - this.Position);
this.Velocity += acceleration * elapsedTime;
```

Before I write tests, I wanna see if we like this approach at all or can think of a better one (or of a reason not to have this code at all).

Secondly, I'd very much like to find better names for the public methods of `GravitationalConstant` - any ideas are much appreciated.


src/SpaceTime/gravity/GravitationalConstant.cs
```diff
@@ -0,0 +1,55 @@
...
+        public Acceleration2 AccelerationTowards(Mass mass, Difference2 differenceToOther)
+        {
+            var distanceSquared = differenceToOther.LengthSquared;
+            var directionVector = differenceToOther / distanceSquared.Sqrt();
+            var acceleration = AccelerationAtDistance(mass, distanceSquared);
+
+            return acceleration * directionVector;
+        }
+
+        public Acceleration AccelerationOnLineTowards(Mass mass, Unit differenceToOther)
```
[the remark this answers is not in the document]
tomrijnbeek: `AccelerationInPlaneTowards` is a possibility. Could always rename the `Acceleration2` method in the future, but that's a breaking change.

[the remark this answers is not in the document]
amulware: I was incorrect. They can have the same name because of the differing parameters.

amulware:
This is ready for review. There is a lot of duplication in the tests though. Do you think I should extract that into a helper setup method? (Also.. why is appveyor failing?)

tests/SpaceTime/GravitationalConstantTests.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            [Property(Arbitrary = new []{typeof(LimitedRangeFloatGenerator)})]
+            public void CalculatesAccelerationCorrectly(
+                float gValue, float massValue, float differenceXValue, float differenceYValue)
+            {
+                var g = new GravitationalConstant(gValue);
+                var mass = new Mass(massValue);
+                var difference = new Difference2(differenceXValue, differenceYValue);
+                var expectedAcceleration = gValue * massValue
+                    * new Vector2(differenceXValue, differenceYValue).Normalized()
```
tomrijnbeek: nit: I would put the vector aspect of this calculation either at the start or the front to leave the numeric part of it intact. I think that would make it more recognisable as `Gm / r^2`.
Also consider extracting a `radiusSquared` variable

(Same below with the unit one)

tests/SpaceTime/GravitationalConstantTests.cs
```diff
@@ -0,0 +1,144 @@
...
+        public class TheAccelerationAtDistanceWithSquaredUnitMethod
+        {
+            [Property(Arbitrary = new []{typeof(LimitedRangeFloatGenerator)})]
+            public void CalculatesAccelerationCorrectly(
+                float gValue, float massValue, float distanceSquaredValueWithSign)
+            {
+                var g = new GravitationalConstant(gValue);
+                var mass = new Mass(massValue);
+                var distanceSquaredValue = Math.Abs(distanceSquaredValueWithSign);
+                var distanceSquared = new Squared<Unit>(Math.Abs(distanceSquaredValue));
```
tomrijnbeek: I don't think you need the second `Math.Abs` here

tests/SpaceTime/GravitationalConstantTests.cs
```diff
@@ -0,0 +1,144 @@
...
+        public class TheAccelerationAtDistanceWithDifference2Method
+        {
+            [Property(Arbitrary = new []{typeof(LimitedRangeFloatGenerator)})]
+            public void CalculatesAccelerationCorrectly(
+                float gValue, float massValue, float differenceXValue, float differenceYValue)
+            {
+                var g = new GravitationalConstant(gValue);
+                var mass = new Mass(massValue);
+                var difference = new Difference2(differenceXValue, differenceYValue);
+                var expectedAcceleration = gValue *massValue
```
tomrijnbeek: nit: missing whitespace

amulware:
This is ready for re-review. All comments above addressed exactly as asked. I don't consider the order of operations a mere nit either, good catch.

PR #155, opened 2019-02-08
Add a project for testing utilities

tomrijnbeek:
This PR adds a new project that can contain testing utilities. These testing utilities are meant to be used to make testing easier if you use some of our types (such as Maybe) by abstracting away common checks.

### DIFF
--- a/testing/Class1.cs
+++ b/testing/Class1.cs
@@ -1,0 +1,4 @@
+﻿namespace Bearded.Utilities.Testing
+{
+    public class Class1 { }
+}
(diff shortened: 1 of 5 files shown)


testing/Class1.cs
```diff
@@ -0,0 +1,4 @@
+﻿namespace Bearded.Utilities.Testing
+{
+    public class Class1 { }
```
amulware: New phone, who dis?

tomrijnbeek: I wasn't sure if the project would even build without any classes, and I wanted to keep the project creation PR separate from the implementation of the first helper /shrug

tomrijnbeek:
Now ready for review as new artifact is correctly picked up.

testing/Class1.cs
tomrijnbeek: Or maybe the builder doesn't care...

PR #160, opened 2019-02-08
Convert prefix tree tests to use fluent assertions

amulware:
What it says on the tin. The old tests weren't only less fluent but also used silly constructions instead of smarter assert statements. That caused a lot of warnings which are now gone. Yay!

### DIFF
--- a/tests/Collections/PrefixTrieTests.cs
+++ b/tests/Collections/PrefixTrieTests.cs
@@ -119,13 +119,11 @@ namespace Bearded.Utilities.Tests.Collections
 
             var withPrefixO = trie.AllKeys("o").ToList();
 
-            withPrefixO.Should().HaveCount(1);
-            withPrefixO.Should().Contain("one");
+            withPrefixO.Should().BeEquivalentTo(new[] { "one" });
 
             var withPrefixTh = trie.AllKeys("th").ToList();
 
-            withPrefixTh.Should().HaveCount(2);
-            withPrefixTh.Should().Contain(words.Where(w => w.StartsWith("th")));
+            withPrefixTh.Should().BeEquivalentTo(words.Where(w => w.StartsWith("th")));
         }
 
         [Fact]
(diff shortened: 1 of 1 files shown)


tests/Collections/PrefixTrieTests.cs
```diff
@@ -112,16 +109,13 @@ public void TestAllKeys()
 
             var withPrefixO = trie.AllKeys("o").ToList();
 
-            Assert.Equal(1, withPrefixO.Count);
-            Assert.Equal("one", withPrefixO[0]);
+            withPrefixO.Should().HaveCount(1);
```
tomrijnbeek:
```suggestion
            withPrefixO.Should().ContainSingle().Which.Should().Be("one");
```

It combines both the conditions in one line. Alternatively:

```suggestion
            withPrefixO.Should().ContainSingle(elmt => elmt == "one");
```

Or

```suggestion
            // the array wrapper is unnecessary, but more explicit
            withPrefixO.Should().BeEquivalentTo(new [] { "one" });
```

tests/Collections/PrefixTrieTests.cs
```diff
@@ -112,16 +109,13 @@ public void TestAllKeys()
...
 
             var withPrefixTh = trie.AllKeys("th").ToList();
 
-            Assert.Equal(2, withPrefixTh.Count);
-            foreach (var word in words.Where(w => w.StartsWith("th")))
-            {
-                Assert.True(withPrefixTh.Contains(word));
-            }
+            withPrefixTh.Should().HaveCount(2);
+            withPrefixTh.Should().Contain(words.Where(w => w.StartsWith("th")));
```
tomrijnbeek:
```suggestion
            withPrefixTh.Should().BeEquivalentTo(words.Where(w => w.StartsWith("th")));
```

Don't need the `HaveCount` then. Same with the test below.

amulware:
Well spotted. I'm not sure what I think about the explicit params array, but it doesn't feel right without it either. Would be nice if the method was called something else, but what can you do.

PR #165, opened 2019-03-03
Implement Where method for Maybe

tomrijnbeek:
I've had the joy of actually using the equivalent of this on `Optional` two days ago, and it can be super helpful, so figured I should implement it and follow the LINQ naming style in .NET.

Note: I once again opted to not use the `Match` for performance reasons, given that this library is directed at games.

I also updated the tests to follow the nested class style that I used in `MaybeAssertions`. Should probably be consistent and in trying whether this works for me, might as well go all the way :)


tomrijnbeek:
I split the test refactor and `Where` implementation in separate commits. May be easier to review by commit in this case.

PR #166, opened 2019-03-03
Add Random tests

amulware:
This adds tests for our random extensions and StaticRandom class.

The tests are the same for both, using shared code And minimal dummy implementations of the abstract classes containing the actual tests for each of the two cases.

This provides full coverage, but the tests are by no means complete. I check the boundary conditions for all methods, but I do not actually ensure that the same seed reproduced the same sequence for all methods or that all methods produce different numbers given 'sane' bounds.

What is here is reasonably good though (I think), so I think we should merge it. If we want to we can always improve it later.

### DIFF
--- a/Bearded.Utilities.Tests/Core/Random/SharedTests.cs
+++ b/Bearded.Utilities.Tests/Core/Random/SharedTests.cs
@@ -1,0 +1,301 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using FsCheck;
+using FsCheck.Xunit;
+
+namespace Bearded.Utilities.Tests.Core.Random
+{
+    public class SharedTests
+    {
+        private static IEnumerable<T> sequence<T>(int count, Func<T> function)
+            => Enumerable.Range(0, count).Select(_ => function());
+
+        private static List<T> list<T>(int count, Func<T> function)
+            => sequence(count, function).ToList();
+        
+        public abstract class TheSeedMethod : BaseTests.RandomMethodWithoutParameters<int>
+        {
+            [Property]
+            public void LeadsToPredictableResultsWithSameSeed(int seed)
+            {
+                Seed(seed);
+                var sequence1 = list(100, CallMethod);
+
+                Seed(seed);
+                var sequence2 = list(100, CallMethod);
+
+                sequence1.Should().BeEquivalentTo(sequence2,
+                    options => options.WithStrictOrdering());
+            }
+        }
+        
+        public abstract class TheIntMethod : BaseTests.RandomMethodWithoutParameters<int>
+        {
+            [Property]
+            public void ShouldReturnDifferentValues(int seed)
+            {
+                Seed(seed);
+                var first = CallMethod();
+
+                sequence(100, CallMethod)
+                    .Should().Contain(i => i != first);
+            }
+        }
+
+        public abstract class TheIntWithMaxMethod : BaseTests.RandomMethodWithOneParameter<int>
+        {
+            [Property]
+            public void ShouldReturnValueInRange(int seed, PositiveInt max)
+            {
+                ResultOfCallingWith(seed, max.Get)
+                    .Should().BeInRange(0, max.Get);
+            }
+
+            [Property]
+            public void ShouldThrowForNegativeMaximum(int seed, NegativeInt max)
+            {
+                CallingWith(seed, max.Get).Should().Throw<ArgumentException>();
+            }
+
+            [Property]
+            public void ShouldReturnZeroForZeroMaximum(int seed)
+            {
+                ResultOfCallingWith(seed, 0).Should().Be(0);
+            }
+        }
+
+        public abstract class TheIntWithMinAndMaxMethod : BaseTests.RandomMethodWithTwoParameters<int>
+        {
+            [Property]
+            public void ShouldReturnValueInRange(int seed, int min, int max)
+            {
+                if (max < min)
+                    (min, max) = (max, min);
+
+                ResultOfCallingWith(seed, min, max)
+                    .Should().BeInRange(min, max);
+            }
+
+            [Property]
+            public void ShouldThrowIfMaximumLessThanMinimum(int seed, int min, int max)
+            {
+                if (max == min)
+                    max--;
+                if (max > min)
+                    (min, max) = (max, min);
+
+                CallingWith(seed, min, max)
+                    .Should().Throw<ArgumentException>();
+            }
+
+            [Property]
+            public void ShouldReturnMinimumIfMinimumEqualsMaximum(int seed, int minMax)
+            {
+                ResultOfCallingWith(seed, minMax, minMax)
+                    .Should().Be(minMax);
+            }
+        }
+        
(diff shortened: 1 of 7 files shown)


Bearded.Utilities.Tests/Core/Random/SharedTests.cs
```diff
@@ -0,0 +1,441 @@
...
+                    .Should().Contain(i => i != first);
+            }
+        }
+
+        public abstract class TheLongWithMaxMethod : BaseTests.RandomMethodWithOneParameter<long>
+        {
+            [Property]
+            public void ShouldReturnValueInRange(int seed, long max)
+            {
+                max = Math.Max(1, max);
```
tomrijnbeek: `max = max == 0 ? 1 : Math.Abs(max)` would provide wider coverage of test cases for the same number of runs.

Bearded.Utilities.Tests/Core/Random/SharedTests.cs
```diff
@@ -0,0 +1,441 @@
...
+        
+        public abstract class TheFloatMethod : BaseTests.RandomMethodWithoutParameters<float>
+        {
+            [Property]
+            public void ShouldReturnDifferentValues(int seed)
+            {
+                Seed(seed);
+                var first = CallMethod();
+
+                sequence(100, CallMethod)
```
tomrijnbeek: The `100` is a recurring constant. Let's extract it to a `const`?

Bearded.Utilities.Tests/Core/Random/SharedTests.cs
```diff
@@ -0,0 +1,441 @@
...
+        }
+        public abstract class TheDoubleMethod : BaseTests.RandomMethodWithoutParameters<double>
+        {
+            [Property]
+            public void ShouldReturnDifferentValues(int seed)
+            {
+                Seed(seed);
+                var first = CallMethod();
+
+                // ReSharper disable once CompareOfDoublesByEqualityOperator
```
tomrijnbeek: Why disable floats globally for this file, but doubles inline?

Bearded.Utilities.Tests/Core/Random/SharedTests.cs
```diff
@@ -0,0 +1,441 @@
...
+                    : (b2, b1);
+
+                ResultOfCallingWith(seed, b1, b2)
+                    .Should().BeInRange(minimumValue, maximumValue);
+            }
+
+            [Property]
+            public void ShouldReturnBoundIfBoundsAreEqual(int seed, NormalFloat bound)
+            {
+                var b = (float) bound.Get;
```
tomrijnbeek: Should't `b` be a `double` here?

Bearded.Utilities.Tests/Core/Random/SharedTests.cs
```diff
@@ -0,0 +1,441 @@
...
+        public abstract class TheBoolWithParameterMethod : BaseTests.RandomMethodWithOneParameter<bool, double>
+        {
+            [Property]
+            public void ReturnsOnlyFalseForZeroParameter(int seed)
+            {
+                ResultOfCallingWith(seed, 0).Should().Be(false);
+            }
+            
+            [Property]
+            public void ReturnsOnlyTrueForOneParameter(int seed)
```
tomrijnbeek: This reads as if we are giving the function one parameter, rather than the parameter `1`.

amulware:
Regarding test clarity, is it really more helpful to have lots of duplication? Is anyone going to read all that code? It's already A LOT.

tomrijnbeek:
> Regarding test clarity, is it really more helpful to have lots of duplication? Is anyone going to read all that code? It's already A LOT.

There is always a trade-off between duplication and code clarity. Not having any abstraction at all would be way too much code, yes. On the other hand, this PR seems to make things very complicated. While the test cases show up nicely in the test output, they aren't really clear from the code. You have to dig through quite a few classes to find the actual implementation. I already find that problematic, because if tests can't be verified as clearly correct, I don't think they're that valuable.

There is a potentially bigger problem though: maintainability. How easy is it to make changes to these tests, or add new tests, when changes are made to the implementation? The complexity of these tests encourages running the tests, seeing which ones fails, and making changes until they pass. Over time, I wonder if these tests will be fully understood by anyone...

If you believe that if in one year from now, you could make changes to these tests with confidence, without having to spend a long time understanding the code, then I'm happy to approve this PR. I just have a nagging feeling that this approach is overly complex.

Bearded.Utilities.Tests/Core/Random/SharedTests.cs
amulware: And throw when `max == long.MinValue`. I don't think it's worth it.

Bearded.Utilities.Tests/Core/Random/SharedTests.cs
amulware: I think it's clear in context. If you insist, suggestions as welcome, but then other methods should probably be renamed too for consistency.

amulware:
I will not refactor this because it's a lot of work. I see your point, and I see my own, and I don't know what's best in the long run. Time will tell.

amulware:
I made the requested changes and/or commented on the comments.

Bearded.Utilities.Tests/Core/Random/SharedTests.cs
tomrijnbeek: I thought the idea of this complicated test structure was so that the tested assertions were clear from the name of the test, which seems to clash with a reasoning that it's clear from context.

Up to you if you think there's anything clearer (you know I dislike the whole sentence-naming anyway), but the alternatives that come to mind are:

* ReturnsOnlyTrueForOne
* ReturnsOnlyTrueForParameterOne
* ReturnsOnlyTrueWhenCalledWithOne
* ReturnsOnlyTrueWhenProbabilityIsOne
* ReturnsOnlyTrueForProbabilityOne

amulware: If you include the context, i.e. the class name, in the test name, I think it's quite clear.

In either case, there's a lot of different options, but unless we agree to some format I'd just leave it this way here for consistency within the same file.

tomrijnbeek:
Of course this is all speaking hypothetically, but personally, I am not very likely to update/expand these tests when new behaviour is added, because it's not a straightforward "navigate to tests, add new tests". I don't want to have to learn a mini piece of architecture spanning multiple classes each time I need to touch the tests. So I don't expect this will really work out in the long run. I'm happy to be proven wrong, and I understand that refactoring this is a lot of work, so I won't block this PR on that. I just hope this isn't the prelude to adding complicated test structures across the board.

Bearded.Utilities.Tests/Core/Random/SharedTests.cs
tomrijnbeek: Even in a full sentence I don't see how the "one" isn't interpreted as a cardinal number in this case. Maybe it's a gold dress, blue dress thing where you can only see one or the other, but I can only find the name of this test misleading. I think just removing the "Parameter" bit takes away both of the bloat of the test name, and removes this confusion, but I won't insist further, since I'll be unlikely to work with these tests myself in the future.

PR #167, opened 2019-05-18
🎯 Add .NET Standard targeting

tomrijnbeek:
One caveat: until we update to OpenTK 4.0.0 (which includes .NET Standard support), this won't be truly cross-platform. Once we do though... 😏


tomrijnbeek:
Ready for review.

tomrijnbeek:
Mostly because it's quite an interface breaking change. That being said,
it's not the first one we would break. If you want me to delete the code, I
will.

amulware:
I would like you to remove it yeah. We don't really care about breaking changes that much.

tomrijnbeek:
Oh, I know why I didn't delete it. Because in the .NET Framework it isn't included before .NET 4.8, and we support as low as .NET 4.6 right now. I don't even have .NET 4.8 installed, so I am not sure if we want to make this change.

PR #171, opened 2019-06-23
🍯 Add single parameter Maybe.Match

amulware:
I think it should be called Match because it's also semantically an overload of the existing Match method.

### DIFF
--- a/Bearded.Utilities.Tests/Core/MaybeTests.cs
+++ b/Bearded.Utilities.Tests/Core/MaybeTests.cs
@@ -5,6 +5,7 @@ using Xunit.Sdk;
 
 namespace Bearded.Utilities.Tests
 {
+    [SuppressMessage("ReSharper", "ArgumentsStyleAnonymousFunction")]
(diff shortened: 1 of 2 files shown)


amulware:
I'm not sure what collision you're referring to. The test class name? That seems like a very very minor issue to me. There is no actual ambiguity when using these methods.

Using pattern matching to consume Maybe doesn't defeat the point, that's just the functional approach. Our methods are object-oriented.

And yes, non-void `Match` is a perfectly valid use case, and of course, the functional thing to do. I'm happy adding that overload for `Match`. Note how it's roughly equivalent to `Select(onValue).ValueOrDefault(onNothing())` except that it's lazy in `onNothing`. We could add a lazy overload for `ValueOrDefault` of course, which might be nice.

I can see the case for `IfJust`, but I don't particularly like `Consume`, mostly on account of matching implying multiple options. But I also like having fewer method names. Especially if we start adding another overload that returns, it seems elegant to me to have only a single consumption method name.

Issue #176, opened 2019-07-22
Make library compliant with recommended open source community standards


tomrijnbeek:
See https://github.com/beardgame/utilities/community (or screenshot below).

![Screenshot 2019-07-22 at 23 37 55](https://user-images.githubusercontent.com/828553/61669958-cdac1b80-acd9-11e9-9ace-68dcfa28fdbf.png)

* [x] Description
* [x] README
* [ ] Code of conduct
* [ ] Contributing guidelines
* [x] License
* [ ] Issue templates
* [ ] Pull request templates

amulware:
Submitted PRs for everything but the code of conduct.

Issue #173, opened 2019-06-30
[Proposal] Release 1.0.0 and keep changelog


tomrijnbeek:
I suggest we start releasing this library on a more regular basis. As a start, I suggest we release 1.0.0 as the first canonical version of this library, and from then onward start keeping a changelog. In particular, we would use the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format, which means we will have an "unreleased" section at the top of our changelog, and each PR would update the changelog with its changes.

In addition, from 1.0.0 onwards, we would also use the [Semantic Versioning](https://semver.org/spec/v2.0.0.html) guidelines for future version numbers.

tomrijnbeek:
To clarify, I am proposing to make these changes across all Bearded libraries, including Bearded.Audio and Bearded.UI, but am using this project as a hub project.

amulware:
I'd like clear definitions of what constitutes a breaking change and what doesn't.

amulware:
I.e. clear specs on the versioning, before I'm happy to go ahead with that.

tomrijnbeek:
I think the semantic versioning guidelines are detailed enough to remove
any ambiguity around versioning. The only think that may not be completely
clear is what constitutes a breaking change, but I believe there is plenty
of information in the official .NET documentation to make that
non-ambiguous.

From the Open source guidance for .NET standard libraries [1], the breaking
change definitions for corefx [2] are linked, which I believe are clear
enough that they constitute as official rules.

[1]
https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes
[2]
https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md

amulware:
Those lists are very long and complicated. :D

tomrijnbeek:
That's what you get if you ask for a definition of a breaking change, no? :P

tomrijnbeek:
My proposal is to:

* Release 1.0.0
* From that point onward, enforce that PRs also update the changelog. [keepachangelog(https://keepachangelog.com/en/1.0.0/) suggest keeping a `CHANGELOG.md` file with an `[Unreleased]` section at the top that needs to be updated as part of each PR.
* Pre-releases can be made from `master` at any time by tagging a commit.
* Full releases constitute of a PR that bumps the version number (if relevant), moves the `[Unreleased]` section of the changelog into a named segment, and a merge into `master`, after which the `merge` commit is tagged.
[...]

PR #179, opened 2019-10-15
📝 Add issue and pull request templates

amulware:
## ✨ What's this?
This PR adds one pull request template (which you see in action here) and several issue templates to help us and others create well-formatted PRs and issues.

### 🔗 Relationships

Part of #176

## 🔍 Why do we want this?
To make it easier for us and others to submit issues and PRs in a well-formatted way. Generally, these sort of templates makes people think a bit more about what they write and help them add more relevant information than they otherwise would which is good for everyone!

## 🏗 How is it done?
By adding markdown files to the `.github` directory. Specifically by adding one pull request template, and several issue templates. When creating a new issue after this is merged, one will be prompted to choose from one of the issue templates which will populate the description and labels of the issue, and all PRs created here on GitHub will automatically have the PR template pre-filled.

All those files also include a bunch of HTML comments which explain each section and what to add there. Ideally, people will replace them with actual content, but if they forget, they won't be rendered in the final issue/PR, so at least we don't have to look at them everywhere (though they'll still show up in our webhook integrations probably..).

### 💥 Breaking changes
Absolutely none. :)

### 🔬 Why not another way?
We could also add multiple pull request templates (it's supported), however, there is no UI to chose from them, and we would have to add `?template=template-file.md` to url manually to use them, which isn't a great solution. I think sticking to a more general PR template is the better tradeoff here by far.

### 🦋 Side effects
Creating issues and PRs will take more time. However, they will also be much easier to read. Worth it!

If people forget to remove the HTML comments, our webhook integration posts will be ugly. However, I think there's too much value in the comments to remove them.

## 💡 Review hints
Read the markdown files, see if you like the layout of this PR, and if you're not sure how the issue templates are gonna look like, check a repo that supports them, or take a look at the documentation here: https://help.github.com/en/articles/about-issue-and-pull-request-templates

### DIFF
--- a/.github/ISSUE_TEMPLATE/b-bug-report.md
+++ b/.github/ISSUE_TEMPLATE/b-bug-report.md
@@ -1,0 +1,23 @@
+---
+name: "🐛 Bug Report"
+about: Found a bug? Click here and let us know!
+labels: bug
+---
+
+<!-- Please give the issue a short, clear title -->
+
+## 🐛 Bug Report
+<!-- Summarise the bug here as clearly and concisely as you can -->
+<!-- If you know where the bug is located, include links to the respective files or even lines of code -->
+
+## Context
+<!-- Give any relevant context, like what platform or version of the library you are using -->
+
+## Expected behavior
+<!-- A clear and concise description of what you expected to happen -->
+
+## To reproduce
+<!-- Clear steps to reproduce the issue -->
+
+## Suggestions
+<!-- Any suggestions on how to investigate, fix or gather more details to resolve the issue faster, and any such details already known -->
--- a/.github/ISSUE_TEMPLATE/b-bug-report.md
+++ b/.github/ISSUE_TEMPLATE/b-bug-report.md
@@ -9,6 +9,7 @@ labels: bug
 ## 🐛 Bug Report
 <!-- Summarise the bug here as clearly and concisely as you can -->
 <!-- If you know where the bug is located, include links to the respective files or even lines of code -->
+<!-- Also include a stack trace or similar information if you have it -->
 
 ## Context
 <!-- Give any relevant context, like what platform or version of the library you are using -->
--- a/.github/ISSUE_TEMPLATE/c-feature-request.md
+++ b/.github/ISSUE_TEMPLATE/c-feature-request.md
@@ -10,7 +10,7 @@ labels: feature
 <!-- Summarise the feature you would like to see as clearly and concisely as you can -->
 
 ## Context
-<!-- Tell us what this feature would be useful for; who would use it, for what purpose? -->
+<!-- Tell us what this feature would be useful for: who would use it, for what purpose? -->
 
 ## Scope
 <!-- List various pieces of functionality you would like this feature to have. This could be a list of concrete functions, or just some bullet points explaining the various things that can be done with it -->
--- a/.github/ISSUE_TEMPLATE/c-feature-request.md
+++ b/.github/ISSUE_TEMPLATE/c-feature-request.md
@@ -14,3 +14,4 @@ labels: feature
 
 ## Scope
 <!-- List various pieces of functionality you would like this feature to have. This could be a list of concrete functions, or just some bullet points explaining the various things that can be done with it -->
+<!-- Also consider telling us what might be too much: what functionality comes to mind that is not needed? -->
--- a/.github/ISSUE_TEMPLATE/c-feature-request.md
+++ b/.github/ISSUE_TEMPLATE/c-feature-request.md
@@ -7,7 +7,7 @@ labels: feature
 <!-- Please give the issue a short, clear title -->
 
 ## ✨ Feature Request
-<!-- Summarise the feature you would like to see here as clearly and concisely as you can -->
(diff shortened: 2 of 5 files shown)


tomrijnbeek:
I have no immediate concerns with this PR. I am however curious as to the source of the template. Right now what is missing from the PR description is why the templates were chosen as they are (which isn’t trivial), e.g. whether they were based on an example. Not that I see anything wrong with them, but I wonder if there’s any best practices around these that we should try to make sure to apply.

amulware:
The templates are based on my experience of what works/is useful, filtered by what I could come up with/remember at the moment.

.github/ISSUE_TEMPLATE/b-bug-report.md
```diff
@@ -0,0 +1,23 @@
+---
+name: "🐛 Bug Report"
+about: Found a bug? Click here and let us know!
+labels: bug
+---
+
+<!-- Please give the issue a short, clear title -->
+
+## 🐛 Bug Report
+<!-- Summarise the bug here as clearly and concisely as you can -->
```
tomrijnbeek: nit: leave empty line below markdown header. You do this for the question template

.github/ISSUE_TEMPLATE/b-bug-report.md
```diff
@@ -0,0 +1,23 @@
...
+<!-- If you know where the bug is located, include links to the respective files or even lines of code -->
```
tomrijnbeek: We should also ask for a stack trace if relevant.

.github/ISSUE_TEMPLATE/c-feature-request.md
```diff
@@ -0,0 +1,16 @@
+---
+name: "✨ Feature Request"
+about: Would you like to ask for or discuss adding a new feature? Let us know!
+labels: feature
+---
+
+<!-- Please give the issue a short, clear title -->
+
+## ✨ Feature Request
+<!-- Summarise the feature you would like to see here as clearly and concisely as you can -->
```
tomrijnbeek: very nit: you don’t say “here” in the other descriptions

.github/ISSUE_TEMPLATE/c-feature-request.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Context
+<!-- Tell us what this feature would be useful for; who would use it, for what purpose? -->
```
tomrijnbeek: very nit: incorrect use of semicolon. Should be a colon or full stop.

.github/ISSUE_TEMPLATE/c-feature-request.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Scope
+<!-- List various pieces of functionality you would like this feature to have. This could be a list of concrete functions, or just some bullet points explaining the various things that can be done with it -->
```
tomrijnbeek: Do we also ask to mention things specifically out of scope?

.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+
+Closes #
+
+## 🔍 Why do we want this?
+<!-- Describe the reason for making this change -->
+
+## 🏗 How is it done?
+<!-- Describe how the changes are implemented, list the different parts the changes consist of if it's more than one -->
+
+### 💥 Breaking changes
```
tomrijnbeek: While it’s good to see breaking changes documented as part of a PR, I actually think it would be better documented in a changelog inside the repository (would also be visible in the review), since breaking changes should be documented for releases. See #173

tomrijnbeek:
General comment about this PR: the issue template folder is allcaps, but the pull request folder is not. Is this intentional?

.github/ISSUE_TEMPLATE/b-bug-report.md
amulware: I removed the empty line in the question template since it was the only place that had it. I see no reason for needing it.

.github/pull_request_template.md
amulware: No harm in having this here, and who knows when the other thing gets implemented consistently.

amulware:
RE capitalisation: I believe the all-caps folder name to be needed (it may not be), and there is no pull request folder.

Addressed all comments, good for re-review afaik.

.github/ISSUE_TEMPLATE/b-bug-report.md
tomrijnbeek: Markdownlint [1] suggests blank lines around the header on both sides (which is what I personally use myself), but doesn't provide a particularly strong argument. I would prefer the blank lines, but this is mostly for aesthetic reasons, so it is not a super strong opinion.

[1] https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md022---headings-should-be-surrounded-by-blank-lines

amulware: Under normal circumstances, I would empty lines but I think in the case of these comments it's fine. (I would also leave two empty lines above ## (h2) for better separation, but I don't think that does any good here.

PR #180, opened 2019-10-15
📝 Add contribution guidelines

amulware:
## ✨ What's this?
This adds relatively basic and straight forward contribution guidelines. I'm sure they can be improved much in the future, but I figured it's a decent start.

### 🔗 Relationships

Part of #176 

Requires #179 for '[Open an issue!](https://github.com/beardgame/utilities/issues/new/choose)' link to work properly

## 🔍 Why do we want this?
So that people unfamiliar with us/the repo feel welcome to create issues and PRs and give them a few simple pointers of how to go about it.

## 🏗 How is it done?
Adding a `CONTRIBUTING.md` to the repo root.

### 💥 Breaking changes
None.

### 🔬 Why not another way?
This is the only way.

### 🦋 Side effects
There's a chance this will lead to people making issues/PRs which results in work for us. Oh no.

## 💡 Review hints
Read the file, tell me what you think.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,8 @@ In principle this as easy as forking the repository, making some changes and [op
 
 If you want to know ahead of time if we'd welcome a certain feature or change, you could always create an issue where you propose that change and mention that you'd like to implement it yourself. We'll let you know what we think as soon as we can. The main reason we might decline code is if we feel it doesn't fit with the purpose of our libraries, but we're very open to suggestions and good arguments.
 
+You can also browse the existing issues for inspiration. You can pick up anything, but we also have the [`good-first-issue` label](https://github.com/beardgame/utilities/labels/good-first-issue) to indicate which issues are easy to pick up for someone new to the code base.
+
 When writing code for Bearded.Utilities, keep the following things in mind:
 
 - We like pull requests to be as small as possible, but as large as necessary, to make them as easy to review as possible
@@ -31,6 +33,7 @@ When writing code for Bearded.Utilities, keep the following things in mind:
 - We like modern C# features, type-safety and care about performance
 - We like tests
 
+Also remember to assign yourself to an issue if you start working on it.
 
 ## General tips and tricks
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,39 @@
+# Contributing to Bearded.Utilities
+
+You want to help us improve our little library? ~~That's awesome!~~ You're awesome!
+
+Whether you have a feature request, found a bug, want to add some code of your own or even just have a question - all of these are very welcome.
+
+Our library is pretty small, so let's try and go easy on the rules. :)
+
+
+## The basics
+
+- **Have a question?** [Open an issue!](https://github.com/beardgame/utilities/issues/new/choose)
+- **Found a bug?** [Open an issue!](https://github.com/beardgame/utilities/issues/new/choose)
+- **Would you like to request a feature, change or improvement?** [Open an issue!](https://github.com/beardgame/utilities/issues/new/choose)
+
+You could also always check out list of issues and see if there's any discussions you want to share your thoughts on.
+
+
+## You'd actually like to make changes yourself?
+
+Wow, that's even better!
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+
+If you want to know ahead of time if we'd welcome a certain feature or change, you could always create an issue where you propose that change and mention that you'd like to implement it yourself. We'll let you know what we think as soon as we can. The main reason we might decline code is if we feel it doesn't fit with the purpose of our libraries, but we're very open to suggestions and good arguments.
+
+When writing code for Bearded.Utilities, keep the following things in mind:
+
+- We like pull requests to be as small as possible, but as large as necessary, to make them as easy to review as possible
+- We like organised, consistent, clean, readable and maintainable code, see what's already there for examples (or submit a change if you think you can improve it!)
+- We like modern C# features, type-safety and care about performance
+- We like tests
+
```
tomrijnbeek: What is somewhat missing right now is how to jump on issues that already exist, if a potential contributor wants to contribute, but doesn’t have a specific feature or bug in mind (e.g. because they want a shiny hacktoberfest t-shirt).

Many repos work with something like a “good first issue” label to promote that, though it may be overkill for this library.

Not a blocker to add a label, but at the very least worth mentioning that any unassigned issue is fair game, though they should probably always assign themselves if they start.

CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+You could also always check out list of issues and see if there's any discussions you want to share your thoughts on.
+
+
+## You'd actually like to make changes yourself?
+
+Wow, that's even better!
+
+In principle this as easy as forking the repository, making some changes and [opening a pull request](https://github.com/beardgame/utilities/compare). From there, we'll review your PR as soon as we can manage, ask questions or request changes if necessary, and once we're happy merge it into our main branch. It's that easy!
+
+If you want to know ahead of time if we'd welcome a certain feature or change, you could always create an issue where you propose that change and mention that you'd like to implement it yourself. We'll let you know what we think as soon as we can. The main reason we might decline code is if we feel it doesn't fit with the purpose of our libraries, but we're very open to suggestions and good arguments.
```
tomrijnbeek: While the probability of duplicate work with the amount of traffic currently is minimal, I am wondering whether we should ask people to always assign themselves to an issue when they start work on something. 

OpenTK works with draft PRs, which I think isn’t a bad idea either, since it is as simple as pushing a branch. However, it works badly with the PR template since the author will have to go back and update the initial post, which as a reviewer is also not gonna be great, especially since it adds a bunch of conversational noise that should maybe be on a FR issue instead.

All in all, maybe having nothing is fine, but in most cases we should recommend an assigned issue for all in flight work.

(Not a blocker for PR approval)

CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+- **Would you like to request a feature, change or improvement?** [Open an issue!](https://github.com/beardgame/utilities/issues/new/choose)
+
+You could also always check out list of issues and see if there's any discussions you want to share your thoughts on.
+
+
+## You'd actually like to make changes yourself?
+
+Wow, that's even better!
+
+In principle this as easy as forking the repository, making some changes and [opening a pull request](https://github.com/beardgame/utilities/compare). From there, we'll review your PR as soon as we can manage, ask questions or request changes if necessary, and once we're happy merge it into our main branch. It's that easy!
```
tomrijnbeek: Who is “we”? In particular, will they have to assign a reviewer? And if so, how to pick one?

CONTRIBUTING.md
amulware: I added the label, will mention it here and also mention issue assignments.

CONTRIBUTING.md
amulware: Will mention issue assignments, I think that's good enough. I'm not a big fan of draft PRs overall, or at least not of mandatory ones since it requires pushing before I've finished my work, which in the case of small issues is an additional effort for little benefit. For bigger changes, sure, makes sense to use them to get feedback, but it's not like we have so much traffic coming through here that it matters for filtering purposes.

CONTRIBUTING.md
amulware: I think it's reasonably clear that 'we' means 'maintainers'/'the people working on/owning this library'. I don't see why people have to pick one or what it would add if they did. We can figure out between ourselves how to handle that when/if it happens. If it ever becomes a problem, we can fix it then.

PR #181, opened 2019-10-22
🚰 Make deletable object list tests use fluent assertions

### LINKED ISSUE #145: Use fluent assertions for DeletableObjectListTests
Depends on #137 and #136

amulware:
✨ What's this?
This changes the `DeletableObjectListTests` to use 100% fluent assertions.

🔗 Relationships
Closes #145

🔍 Why do we want this?
Because we like fluent assertions. They read nicer.

🏗 How is it done?
I replaced all the assertions. Most were trivial but comparing sequences and actually making it check that the objects are exactly the same and not just equivalent objects was not. I added a helper for that at the very top that is used throughout the file.

💥 Breaking changes
None.

🔬 Why not another way?
I was thinking of making the sequence comparison more formal with an extension method or even adding it to our testing framework, but that's maybe overkill unless we start using it in other places.

🦋 Side effects
I also did some mild cleanup (second commit). Mostly to get rid of warnings and make intentions in the code clearer.

💡 Review hints
Not much you can do but check all the changes. If you want to verify that the helper lambda works correctly, you can find some of the simpler tests and shuffle the expected list of items and you'll see the test fail (most likely, it's random).

### DIFF
--- a/Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
+++ b/Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
@@ -4,6 +4,8 @@ using System.Collections.Generic;
 using System.Linq;
 using Bearded.Utilities.Collections;
 using Bearded.Utilities.Linq;
+using FluentAssertions;
+using FluentAssertions.Equivalency;
 using FsCheck;
 using FsCheck.Xunit;
 using Xunit;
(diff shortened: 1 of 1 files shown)


amulware:
Agreed, generally, I think fluent assertions isn't the most intuitive around collections. To be fair though, it's difficult to be clear here without very long method names, since there are so many different things someone might want to assert.

PR #182, opened 2019-11-06
Add reverse enumeration of deletable object list

### LINKED ISSUE #62: Allow backwards enumeration of DeletableObjectList
I have found the collection very useful in all sorts of places, most recently in a custom event-subscription system. For that I may want to enumerate the list in reverse, which can be done almost trivially by adding another Enumerator.

While the current enumerator also enumerates items that are added during enumeration, this new one would not, but that seems a minor issue, or maybe even desired behaviour.

I propose adding the following property:

```
IEnumerable<T> DeletableObjectList<T>.Reversed { get; }
```

The list could keep a no-data object internally which is returned with this property and implements GetEnumerator() correctly. While a bit of a shame in terms of memory, this will make sure there is no unnecessary pressure on the GC.

paulcscharf:
## ✨ What's this?
This implements efficient reverse enumeration of a deletable object list. Items deleted during enumeration are still ignored, and the enumeration has constant time overhead, instead of linear (like Linq's Reverse() would be).

Most importantly, items added during enumeration are not enumerated this way, which is the key use case.

### 🔗 Relationships

Closes #62

## 🔍 Why do we want this?
Sometimes we don't want to enumerate items we add during enumeration until the next time. This allows for that (assuming we don't mind the inverted order of course).

## 🏗 How is it done?
I added a Reversed property that returns a special cached enumerable wrapper which can create a special enumerator implementing this behaviour.

### 💥 Breaking changes
None.

## 💡 Review hints
Mostly check that the new tests make sense. They are mostly adapted copies of the existing enumeration tests.

### DIFF
--- a/Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
+++ b/Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
@@ -25,7 +25,7 @@ namespace Bearded.Utilities.Tests.Collections
         private static List<TestDeletable> getDeletables(int itemsToAdd)
             => Enumerable.Range(0, itemsToAdd).Select(i => new TestDeletable()).ToList();
         
-        private static (DeletableObjectList<TestDeletable> List, List<TestDeletable> Items)
+        private static (DeletableObjectList<TestDeletable> List, IList<TestDeletable> Items)
             createPopulatedList(int itemsToAdd)
         {
             var list = new DeletableObjectList<TestDeletable>();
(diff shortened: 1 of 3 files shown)


Bearded.Utilities/Collections/ReversedDeletableObjectListEnumerator.cs
```diff
@@ -0,0 +1,62 @@
...
+        private int i;
+
+        object System.Collections.IEnumerator.Current => Current;
+        public T Current { get; private set; }
+
+        public ReversedDeletableObjectListEnumerator(DeletableObjectList<T> deletableObjectList, List<T> list)
+        {
+            this.deletableObjectList = deletableObjectList;
+            this.list = list;
+            i = list.Count;
```
tomrijnbeek: very nitpicky: should we just call `Reset()` here instead of setting `i`? This helps making sure that the resetting code between the constructor and the reset function remains in sync.

Bearded.Utilities/Collections/DeletableObjectList.cs
```diff
@@ -106,8 +107,18 @@ public void Clear()
 
         #region Enumeration
 
+        private IEnumerable<T> reversed;
+
+        /// <summary>
+        /// Returns an enumerable that allows for enumerating the collection backwards.
+        /// Items added during backwards enumeration will not be enumerated.
+        /// Unlike a call to Reverse(), this property works in constant time and upcoming items deleted
```
tomrijnbeek: nit: make `Reverse` an actual code link.

Bearded.Utilities/Collections/DeletableObjectList.cs
```diff
@@ -176,5 +187,18 @@ private void compact()
 
         #endregion
 
+        sealed class ReverseEnumerable : IEnumerable<T>
```
tomrijnbeek: nit: mark `private` explicitly

Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
```diff
@@ -473,6 +473,106 @@ public void WorksNonGenerically()
...
+                var (list, items) = createPopulatedList(20);
+                var enumeratedItems = new List<TestDeletable>();
+
+                // ReSharper disable once GenericEnumeratorNotDisposed
+                var enumerator = list.Reversed.GetEnumerator();
+                enumerator.MoveNext();
+                enumerator.MoveNext();
+                enumerator.MoveNext();
+                
+                list.Remove(items[19]);
```
tomrijnbeek: You are testing two different behaviours in a single test. Tests should generally only test a single behaviour each, to make it clear what is broken when a test fails.

Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
```diff
@@ -473,6 +473,106 @@ public void WorksNonGenerically()
...
+                var enumeratedItems = new List<TestDeletable>();
+
+                foreach (var item in list.Reversed)
+                {
+                    enumeratedItems.Add(item);
+                    list.First().Deleted = true;
+                }
+
+                enumeratedItems.Should().BeEquivalentTo(items.Reverse().Take(10), withExactSameItems);
+            }
```
tomrijnbeek: This test's correctness relies quite a lot on getting logic and control structures right. It took me quite a long time to actually understand why this test is correct. That scares me a bit. The guarantee of correctness of the entire implementation relies on the tests being correct as well. That is straightforward if you use some simple examples, but as soon as you start using any control structures, suddenly I need to spend several minutes going through this method line by line to make sure it indeed behaves as we expect.

This is the case for most test methods added here. Some of them purposefully try to be a bit smart, but I don't think that adds to the value of the tests. In this particular case for example, I don't really see why the following test provides any less coverage:

```cs
[Fact]
public void SkipsItemsDeletedAheadOfEnumeration()
{
  // Arrange
  var (list, items) = createPopulatedList(20);
  list.First().Deleted = true;
 
  // Act
  var enumeratedItems = list.Reversed.ToList();

  // Assert
  enumeratedItems.Should().BeEquivalentTo(items.Reverse().Take(19), withExactSameItems);
}
```

If you specifically want to test that it skips enumerated items after the enumerator has been created, I would also make that more explicit:

```cs
[Fact]
public void SkipsItemsDeletedAheadOfEnumeration()
{
  // Arrange
  var (list, items) = createPopulatedList(20);
  var enumerator = list.Reversed.GetEnumerator();
  var enumeratedItems = new List<TestDeletable> { enumerator.Current };
  enumerator.MoveNext(); // make sure the enumerator has started doing things
  list.First().Deleted = true;
 
  // Act
  do
  {
    enumeratedItems.Add(enumerator.Current);
  } while (enumerator.MoveNext());

  // Assert
  enumeratedItems.Should().BeEquivalentTo(items.Reverse().Take(19), withExactSameItems);
}
```

Note that I also marked the three blocks of tests (arrange-act-assert) explicitly. See for example [this documentation](https://docs.microsoft.com/en-us/visualstudio/test/unit-test-basics?view=vs-2019#write-your-tests) on unit testing basics explaining this AAA structure, or [this blog](https://defragdev.com/blog/?p=783) explaining some of the benefits (tl;dr, it helps identifying code smells much easier).

Even these examples I gave already contain more logic and control structures than I'm comfortable with in unit tests (especially the second), but at least their use is minimised, making a test much more easier to verify on correctness, without sacrificing any test coverage in this particular case.

Issue #184, opened 2020-06-01
Enforce nullable reference types


tomrijnbeek:
## 🆙 Improvement proposal
Null checking has been enabled. This means all nullability warnings need to be fixed. Many cases are pretty straightforward (null is expected, or tests verify null inputs throw exceptions), but in some cases the nullability is undesirable.

All these warnings need to be cleaned up one way or another, before enabling full nullability enforcement.

## Context
Null bad. Maybe better.

PR #185, opened 2020-06-01
⬆ Enable C#8 language features and nullability warnings

tomrijnbeek:
## ✨ What's this?
This PR switches the language support for all projects to C# 8. It also enables warnings for nullable reference types. It does not enforce them yet, as several classes need cleanup beyond the scope of a single PR.

### 🔗 Relationships
See #184 for follow-up task of moving to full nullability enforcement.

## 🔍 Why do we want this?
C# 8 leads to nicer, more readable code. Nullable reference type enforcement discourages the use of null, which in turn improves code stability.

## 🏗 How is it done?
The project files are updated. The other changes were either simple search and replaces, or were done with IDE support.

The big manual change was the `DeletableObjectListTests` file. It used `IEnumerable<object[]>` for parameterised tests, but nullability doesn't like that. There is actually a typesafe way of passing in data using `TheoryData`, so I just switched to that instead.

### 💥 Breaking changes
None.

### 🔬 Why not another way?
N/A

### 🦋 Side effects
We may see our return types in the future be moved from non-nullable to nullable. However, we're not in a worse state now than we were before.

## 💡 Review hints
Each of the commits is a simple change (by design), or a search and replace, so reviewing commit by commit may be easier than reviewing all changes together.

### DIFF
--- a/Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
+++ b/Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
@@ -18,13 +18,11 @@ namespace Bearded.Utilities.Tests.Collections
         private static readonly Func<EquivalencyAssertionOptions<TestDeletable>, EquivalencyAssertionOptions<TestDeletable>>
             withExactSameItems = o => o.WithStrictOrdering().ComparingByValue<TestDeletable>();
 
-        public static IEnumerable<object[]> PositiveCounts =>
-            new [] {1, 2, 3, 4, 5, 7, 10, 13, 37, 42, 1337}
-                .Select(i => new object[] { i });
-        
+        public static TheoryData<int> PositiveCounts => new TheoryData<int> {1, 2, 3, 4, 5, 7, 10, 13, 37, 42, 1337};
+
         private static List<TestDeletable> getDeletables(int itemsToAdd)
             => Enumerable.Range(0, itemsToAdd).Select(i => new TestDeletable()).ToList();
-        
+
         private static (DeletableObjectList<TestDeletable> List, List<TestDeletable> Items)
             createPopulatedList(int itemsToAdd)
         {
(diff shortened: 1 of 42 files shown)


Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
```diff
@@ -79,13 +77,20 @@ public void ThrowsForNegativeValues(int count)
...
+
         public abstract class MethodThatDoesNotThrowsWhenEnumeratingTests
         {
             protected abstract void CallMethod(DeletableObjectList<TestDeletable> list);
 
-            public static IEnumerable<object[]> IndicesFromZeroToNineteen =
-                Enumerable.Range(0, 20).Select(i => new object[] {i});
+            public static TheoryData<int> IndicesFromZeroToNineteen = new TheoryData<int>();
+
+            static MethodThatDoesNotThrowsWhenEnumeratingTests()
```
amulware, line 87: Use of this static constructor confused me for a moment. I’d rather use a collection initialiser above if you don’t mind, and turn it into a read only property as well (not sure why it wasn’t before?).

Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
```diff
@@ -18,13 +18,11 @@ public class DeletableObjectListTests
         private static readonly Func<EquivalencyAssertionOptions<TestDeletable>, EquivalencyAssertionOptions<TestDeletable>>
             withExactSameItems = o => o.WithStrictOrdering().ComparingByValue<TestDeletable>();
 
-        public static IEnumerable<object[]> PositiveCounts =>
-            new [] {1, 2, 3, 4, 5, 7, 10, 13, 37, 42, 1337}
-                .Select(i => new object[] { i });
-        
+        public static TheoryData<int> PositiveCounts => new TheoryData<int> {1, 2, 3, 4, 5, 7, 10, 13, 37, 42, 1337};
```
amulware, line 21: Why was this change necessary?

amulware: Answered above.

Bearded.Utilities.Tests/Collections/DeletableObjectListTests.cs
tomrijnbeek: Totally get that. I was on the fence too, but I forgot about making it a self-contained getter.

PR #191, opened 2020-10-04
✨ Implement Result monad

tomrijnbeek:
## ✨ What's this?
This PR adds a `Result` monad, implemented much in the same fashion as the existing `Maybe` monad.

### 🔗 Relationships
N/A

## 🔍 Why do we want this?
The `Result` monad allows methods to return either a success or an error state without having to rely on exceptions.

## 🏗 How is it done?
The `Result` monad follows the patterns of the `Maybe` monad pretty closely.

A readonly generic `Result` struct represents the actual monad. A static non-generic `Result` is added for easier factory methods. Finally, generic `Success` and `Failure` structs are added to allow easy upcasting to a `Result`, similar to how `NothingMaybe` allows upcasting to `Maybe`. This means code can do `Result.Success(myResult)` and as long as the compiler can figure out the error from context (for example the return value of your method), it should work. Note that this isn't infallible, as you can see in the tests.

### 💥 Breaking changes
N/A

### 🔬 Why not another way?
Structs are good?

### 🦋 Side effects
I made the choice of putting the `Result` monad into a new `Monads` namespace instead of `Core`. `Maybe` should probably move there as well, which will be a future breaking change.

## 💡 Review hints
N/A

### DIFF
--- a/Bearded.Utilities/Core/Maybe.cs
+++ b/Bearded.Utilities/Core/Maybe.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using Bearded.Utilities.Monads;
 
 namespace Bearded.Utilities
 {
@@ -20,6 +21,14 @@ namespace Bearded.Utilities
         internal static Maybe<T> Just(T value) => new Maybe<T>(value);
 
         public T ValueOrDefault(T @default) => hasValue ? value : @default;
+
+        public T ValueOrDefault(Func<T> defaultProvider) => hasValue ? value : defaultProvider();
+
+        public Result<T, TError> ValueOrFailure<TError>(TError error) =>
+            hasValue ? (Result<T, TError>) Result.Success(value) : Result.Failure(error);
+
+        public Result<T, TError> ValueOrFailure<TError>(Func<TError> errorProvider) =>
+            hasValue ? (Result<T, TError>) Result.Success(value) : Result.Failure(errorProvider());
 
         public Maybe<TOut> Select<TOut>(Func<T, TOut> selector) =>
             hasValue ? Maybe.Just(selector(value)) : Maybe<TOut>.Nothing;
--- a/Bearded.Utilities/Monads/Result.cs
+++ b/Bearded.Utilities/Monads/Result.cs
@@ -74,13 +74,7 @@ namespace Bearded.Utilities.Monads
         public static bool operator !=(Result<TResult, TError> left, Result<TResult, TError> right) =>
             !left.Equals(right);
 
-        public override int GetHashCode()
-        {
-            var hashCode = isSuccess.GetHashCode();
-            hashCode = (hashCode * 397) ^ EqualityComparer<TResult>.Default.GetHashCode(result);
-            hashCode = (hashCode * 397) ^ EqualityComparer<TError>.Default.GetHashCode(error);
-            return hashCode;
-        }
+        public override int GetHashCode() => HashCode.Combine(isSuccess, result, error);
 
         public override string ToString() => isSuccess ? $"success {result}" : $"error {error}";
 
--- a/Bearded.Utilities/Monads/Result.cs
+++ b/Bearded.Utilities/Monads/Result.cs
@@ -106,7 +106,7 @@ namespace Bearded.Utilities.Monads
 
     public readonly struct Success<T>
     {
-        public readonly T Result;
+        public T Result { get; }
 
         internal Success(T result)
         {
@@ -116,7 +116,7 @@ namespace Bearded.Utilities.Monads
 
     public readonly struct Failure<T>
     {
-        public readonly T Error;
+        public T Error { get; }
 
         internal Failure(T error)
         {
--- a/Bearded.Utilities/Monads/Result.cs
+++ b/Bearded.Utilities/Monads/Result.cs
@@ -1,0 +1,128 @@
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace Bearded.Utilities.Monads
+{
+    public readonly struct Result<TResult, TError> : IEquatable<Result<TResult, TError>>
+    {
+        private readonly bool isSuccess;
+        private readonly TResult result;
+        private readonly TError error;
+
+        private Result(bool isSuccess, TResult result, TError error)
+        {
+            this.isSuccess = isSuccess;
+            this.result = result;
+            this.error = error;
+        }
+
+        public TResult ResultOrDefault(TResult @default) => isSuccess ? result : @default;
+
+        public TResult ResultOrDefault(Func<TResult> defaultProvider) => isSuccess ? result : defaultProvider();
+
+        public TResult ResultOrThrow(Func<TError, Exception> exceptionProvider) =>
+            isSuccess ? result : throw exceptionProvider(error);
+
+        public Result<TOut, TError> Select<TOut>(Func<TResult, TOut> selector) =>
+            isSuccess ? (Result<TOut, TError>) Result.Success(selector(result)) : Result.Failure(error);
+
+        public Result<TOut, TError> SelectMany<TOut>(Func<TResult, Result<TOut, TError>> selector) =>
+            isSuccess ? selector(result) : Result.Failure(error);
+
+        public void Match(Action<TResult> onSuccess)
+        {
+            if (isSuccess)
+            {
+                onSuccess(result);
+            }
+        }
+
+        public void Match(Action<TResult> onSuccess, Action<TError> onFailure)
+        {
+            if (isSuccess)
+            {
+                onSuccess(result);
+            }
+            else
+            {
+                onFailure(error);
+            }
+        }
+
+        public TOut Match<TOut>(Func<TResult, TOut> onSuccess, Func<TError, TOut> onFailure) =>
+            isSuccess ? onSuccess(result) : onFailure(error);
+
+        public bool Equals(Result<TResult, TError> other) =>
+            isSuccess == other.isSuccess
+            && EqualityComparer<TResult>.Default.Equals(result, other.result)
+            && EqualityComparer<TError>.Default.Equals(error, other.error);
+
+        public override bool Equals(object obj) => obj is Result<TResult, TError> other && Equals(other);
+
+        public static bool operator ==(Result<TResult, TError> left, Result<TResult, TError> right) =>
+            left.Equals(right);
+
+        public static bool operator !=(Result<TResult, TError> left, Result<TResult, TError> right) =>
+            !left.Equals(right);
+
+        public override int GetHashCode()
+        {
+            var hashCode = isSuccess.GetHashCode();
+            hashCode = (hashCode * 397) ^ EqualityComparer<TResult>.Default.GetHashCode(result);
+            hashCode = (hashCode * 397) ^ EqualityComparer<TError>.Default.GetHashCode(error);
+            return hashCode;
+        }
+
+        public override string ToString() => isSuccess ? $"success {result}" : $"error {error}";
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static implicit operator Result<TResult, TError>(Success<TResult> success)
+        {
+            return new Result<TResult, TError>(true, success.Result, default);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static implicit operator Result<TResult, TError>(Failure<TError> failure)
+        {
+            return new Result<TResult, TError>(false, default, failure.Error);
+        }
+    }
+
+    public static class Result
+    {
+        public static Result<TResult, TError> Success<TResult, TError>(TResult result) => Success(result);
+
+        public static Success<T> Success<T>(T result) => new Success<T>(result);
+
+        public static Result<TResult, TError> Failure<TResult, TError>(TError error) => Failure(error);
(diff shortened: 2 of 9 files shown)


Bearded.Utilities/Monads/Result.cs
```diff
@@ -0,0 +1,128 @@
...
+        public static TResult ResultOrThrow<TResult, TError>(this Result<TResult, TError> result)
+            where TError : Exception
+        {
+            return result.ResultOrThrow(e => e);
+        }
+    }
+
+    public readonly struct Success<T>
+    {
+        internal readonly T Result;
```
amulware, line 111: Any reason not to expose these publicly?

Bearded.Utilities/Monads/Result.cs
```diff
@@ -0,0 +1,128 @@
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace Bearded.Utilities.Monads
+{
+    public readonly struct Result<TResult, TError> : IEquatable<Result<TResult, TError>>
```
amulware, line 7: What is your reason for making TError generic instead of Exception?

Bearded.Utilities/Monads/Result.cs
```diff
@@ -0,0 +1,128 @@
...
+    {
+        private readonly bool isSuccess;
+        private readonly TResult result;
+        private readonly TError error;
+
+        private Result(bool isSuccess, TResult result, TError error)
```
amulware, line 13: Do I see correctly that the only way of construction is via the Success and Failure casts?

Bearded.Utilities/Monads/Result.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        public void Match(Action<TResult> onSuccess)
+        {
+            if (isSuccess)
+            {
+                onSuccess(result);
+            }
+        }
+
+        public void Match(Action<TResult> onSuccess, Action<TError> onFailure)
```
amulware, line 41: Something that worries my slightly is that default(Result<,>) will have error == null, which could lead to a NullReferenceException in onFailure here and below.

Bearded.Utilities/Monads/Result.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        public TResult ResultOrDefault(TResult @default) => isSuccess ? result : @default;
+
+        public TResult ResultOrDefault(Func<TResult> defaultProvider) => isSuccess ? result : defaultProvider();
+
+        public TResult ResultOrThrow(Func<TError, Exception> exceptionProvider) =>
+            isSuccess ? result : throw exceptionProvider(error);
+
+        public Result<TOut, TError> Select<TOut>(Func<TResult, TOut> selector) =>
+            isSuccess ? (Result<TOut, TError>) Result.Success(selector(result)) : Result.Failure(error);
```
amulware, line 28: Why not (and same below)
```suggestion
            isSuccess ? (Result<TOut, TError>) Result.Success(selector(result)) : this;
```

Bearded.Utilities/Monads/Result.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        public static bool operator !=(Result<TResult, TError> left, Result<TResult, TError> right) =>
+            !left.Equals(right);
+
+        public override int GetHashCode()
+        {
+            var hashCode = isSuccess.GetHashCode();
+            hashCode = (hashCode * 397) ^ EqualityComparer<TResult>.Default.GetHashCode(result);
+            hashCode = (hashCode * 397) ^ EqualityComparer<TError>.Default.GetHashCode(error);
+            return hashCode;
```
amulware, line 74: Should we start using the .NET provided hashcode functionality?

Bearded.Utilities/Monads/Result.cs
```diff
@@ -0,0 +1,128 @@
...
+        private readonly TError error;
+
+        private Result(bool isSuccess, TResult result, TError error)
+        {
+            this.isSuccess = isSuccess;
+            this.result = result;
+            this.error = error;
+        }
+
+        public TResult ResultOrDefault(TResult @default) => isSuccess ? result : @default;
```
amulware, line 20: It's perhaps slightly funny that we provide these easy ways of ignoring the error entirely. In either case, it got me thinking: should there be an easy way to convert between Result and Maybe?

tomrijnbeek: Good idea.

Right now this adds a circular dependency between `Core` and `Monads`, but that will be solved once we move `Result` and `Maybe` to the same namespace.

Bearded.Utilities/Monads/Result.cs
tomrijnbeek: Because the success type isn't the same, and you can't cast `Result<TResult, TError>` to `Result<TOut, TError>`, unless you case it back to a `Failure<TError>` first.

Bearded.Utilities/Monads/Result.cs
tomrijnbeek: That's a good point.

Do you have any suggestions on how to resolve this? I can't think of a good way to avoid.

Bearded.Utilities/Monads/Result.cs
tomrijnbeek: I can't access `System.HashCode` right now. My guess is that is because we're still targeting netframework at `developer` for this PR.

Happy to do a pass over all our code to switch to the .NET provided functionality in one go after our official switchover.

Bearded.Utilities/Monads/Result.cs
tomrijnbeek: Nope, no reason.

Bearded.Utilities/Monads/Result.cs
amulware: > imagine this error type being an HTTP error type

Use an http exception with a status field? But sure, I'm fine leaving this as is. I just note that it is not much more than an Either type with fancy naming then. :)

Bearded.Utilities/Monads/Result.cs
amulware: Maybe<T> has public construction methods on the type itself. I think we should perhaps have them here as well, so people don't have to use the other types? Or do you think we should enforce the 'nice' way (really the way that we think is nicest)?

Bearded.Utilities/Monads/Result.cs
amulware: The only thing we can do is catching default state ourselves and throwing (can't check for null, since someone might want to Result<T, MyError?>, would need to have another flag, and check that all over the place). I'm not sure if that's how we want this to behave? Ideally, structs have meaningful default values... in this case that would only work if we enforce TError to be nullable, which isn't great.

Bearded.Utilities/Core/Maybe.cs
```diff
@@ -21,6 +22,14 @@ private Maybe(T value)
 
         public T ValueOrDefault(T @default) => hasValue ? value : @default;
 
+        public T ValueOrDefault(Func<T> defaultProvider) => hasValue ? value : defaultProvider();
+
+        public Result<T, TError> ValueOrFailure<TError>(TError error) =>
+            hasValue ? (Result<T, TError>) Result.Success(value) : Result.Failure(error);
+
+        public Result<T, TError> ValueOrFailure<TError>(Func<TError> errorProvider) =>
```
amulware, line 30: The name is not ideal since it does something quite different than ValueOrDefault, but I can't think of a better one either.

Bearded.Utilities/Monads/Result.cs
```diff
@@ -24,6 +24,8 @@ private Result(bool isSuccess, TResult result, TError error)
         public TResult ResultOrThrow(Func<TError, Exception> exceptionProvider) =>
             isSuccess ? result : throw exceptionProvider(error);
 
+        public Maybe<TResult> ResultMaybe() => isSuccess ? Maybe.Just(result) : Maybe.Nothing;
```
amulware, line 27: This name bothers me. Should it perhaps be `AsMaybe`? or `MaybeResult`?

Bearded.Utilities/Monads/Result.cs
amulware: Using extension methods would have prevented the circular dependency, however it would require matching which is less efficient, so I agree with this approach.

Bearded.Utilities/Monads/Result.cs
tomrijnbeek: Sure, but then `Result` still needs to be generic in the second parameter, correct? So is the suggestion that we add the `where TError : Exception` type requirement, or am I misinterpreting?

But yes, you're right, it's basically an `Either` with different semantics.

Bearded.Utilities/Monads/Result.cs
tomrijnbeek: Fair enough, added.

Bearded.Utilities/Core/Maybe.cs
tomrijnbeek: What about `AsResult` to mimic the `AsMaybe` on `Result`?

Bearded.Utilities/Monads/Result.cs
tomrijnbeek: Updated now that we have .NET Core

Bearded.Utilities/Core/Maybe.cs
amulware: I see what you mean, but that makes the parameter not read very well/clearly perhaps?

```
someMaybe.AsResult(() => new MyException())
```

 feels pretty weird.

Bearded.Utilities/Monads/Result.cs
amulware: No concrete suggestion is intended, but yes, that generic constraint would be what it would end up as. I'm happy leaving this as is.

Bearded.Utilities/Monads/Result.cs
```diff
@@ -0,0 +1,126 @@
...
+        public static TResult ResultOrThrow<TResult, TError>(this Result<TResult, TError> result)
+            where TError : Exception
+        {
+            return result.ResultOrThrow(e => e);
+        }
+    }
+
+    public readonly struct Success<T>
+    {
+        public readonly T Result;
```
amulware, line 109:
```suggestion
        public T Result { get; }
```

Bearded.Utilities/Monads/Result.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        internal Success(T result)
+        {
+            Result = result;
+        }
+    }
+
+    public readonly struct Failure<T>
+    {
+        public readonly T Error;
```
amulware, line 119:
```suggestion
        public T Error { get; }
```

amulware (changes requested):

amulware:
The small changes requested, and two open discussions above. I see no objective solution for either of them, but let's try to figure those out soon or just go ahead without.

Bearded.Utilities/Core/Maybe.cs
tomrijnbeek: I think my preference here is either `ValueOrFailure` or `AsResult`. I don't think I have a strong preference for either.

Bearded.Utilities/Monads/Result.cs
tomrijnbeek: We could check it everywhere, but what would we do if we found that the result is neither success nor failure? Throw an exception? That doesn't really get us much further. Do we add a third "onDefault" to all match functions? That also seems weird.

My preference would be to leave as-is for the time being, because any solution I can think of complicates the type, without really providing a satisfying solution.

tomrijnbeek:
I have applied the fixes. For the two remaining open discussions:

* If `AsMaybe` is preferred over `ValueOrFailure`, I am happy to change.
* I do not believe there is a good solution for the default `Result` instance as explained above, and it will be up to the caller to make sure `default(TError)` is handled.

Given that this is where we are at, I believe we should merge, changing the name as `ValueOrFailure` if needed.

Bearded.Utilities/Core/Maybe.cs
amulware: Let's leave it as is.

Bearded.Utilities/Monads/Result.cs
amulware: Agreed. Though is there a way for us to mark error as nullable or something? I guess not, given that it can be a valuetype? Perhaps there is an attribute we can use?

tomrijnbeek: Please take a look at the proposed changes based on some added annotations.

Bearded.Utilities/Monads/Result.cs
```diff
@@ -1,16 +1,19 @@
...
     public readonly struct Result<TResult, TError> : IEquatable<Result<TResult, TError>>
     {
         private readonly bool isSuccess;
+        [AllowNull]
         private readonly TResult result;
+        [AllowNull]
         private readonly TError error;
 
-        private Result(bool isSuccess, TResult result, TError error)
+        private Result(bool isSuccess, [AllowNull] TResult result, [AllowNull] TError error)
```
amulware, line 16: Are these necessary? Same with the annotations on the fields? Are they there so we can pass `default` in the static construction methods without issues?

tomrijnbeek:
Yes, that's correct. This was caused by the nullable PR being merged.

amulware (approved):

PR #194, opened 2020-10-12
🏗 Refactor bin packing

amulware:
## ✨ What's this?
This is a medium sized refactoring of the bin packing algorithm to make it easier to read, semantically more correct, and ready for enabling nullability.

### 🔗 Relationships

Ref #184 (this is a necessary first step towards closing this issue)

## 🔍 Why do we want this?
I had some trouble understanding how the algorithm works when I tried making it properly nullable. So I commented the most important parts of it and generally made it more semantic and more readable (I think). Next time we need to make changes to it it should be a lot easier to understand.

### 💥 Breaking changes
~~The public interface now makes nullable return value explicit. This constitutes a breaking change for anyone who has nullable enabled in their code.~~

EDIT: The public interface was changed significantly, and no longer returns null for an empty input collection.

### 🔬 Why not another way?
Didn't need to do this. I think it's a lot better now though.

### 🦋 Side effects
There are some warnings because nullable isn't enabled properly yet. My next PR will address this.

## 💡 Review hints
Read the code, especially the comments if you want to understand the algorithm. I did not significantly change the tests however, so my confidence in these changes is very high.

### DIFF
--- a/Bearded.Utilities.Tests/Algorithms/BinPackingTests.cs
+++ b/Bearded.Utilities.Tests/Algorithms/BinPackingTests.cs
@@ -18,7 +18,8 @@ namespace Bearded.Utilities.Tests.Algorithms
 
             var result = BinPacking.Pack(input);
 
-            Assert.Equal(100, result.Rectangles.Count);
+            Assert.NotNull(result);
+            Assert.Equal(100, result!.Rectangles.Count);
 
             var checkedIds = new HashSet<int>();
 
@@ -43,7 +44,9 @@ namespace Bearded.Utilities.Tests.Algorithms
 
             var result = BinPacking.Pack(input);
 
-            foreach (var r1 in result.Rectangles)
+            Assert.NotNull(result);
+
+            foreach (var r1 in result!.Rectangles)
             {
                 foreach (var r2 in result.Rectangles)
                 {
@@ -70,7 +73,9 @@ namespace Bearded.Utilities.Tests.Algorithms
 
             var result = BinPacking.Pack(input);
 
-            var totalArea = result.Width * result.Height;
+            Assert.NotNull(result);
+
+            var totalArea = result!.Width * result.Height;
 
             Assert.Equal(totalArea, result.Area);
 
@@ -78,7 +83,7 @@ namespace Bearded.Utilities.Tests.Algorithms
 
             var maxX = 0;
             var maxY = 0;
-            
+
             foreach (var r in result.Rectangles)
             {
                 var area = r.Width * r.Height;
@@ -107,14 +112,16 @@ namespace Bearded.Utilities.Tests.Algorithms
             var resultSingle = BinPacking.Pack(input, false);
             var resultMultiple = BinPacking.Pack(input, true);
 
-            Assert.True(resultSingle.Filled <= resultMultiple.Filled);
+            Assert.NotNull(resultSingle);
+            Assert.NotNull(resultMultiple);
+            Assert.True(resultSingle!.Filled <= resultMultiple!.Filled);
         }
 
         [Fact]
         public void TestPackedRectangles_EmptyInputReturnsNull()
         {
             var result = BinPacking.Pack(new List<BinPacking.Rectangle<int>>());
-            
+
             Assert.Null(result);
         }
     }
--- a/Bearded.Utilities.Tests/Algorithms/BinPackingTests.cs
+++ b/Bearded.Utilities.Tests/Algorithms/BinPackingTests.cs
@@ -118,11 +118,17 @@ namespace Bearded.Utilities.Tests.Algorithms
         }
 
         [Fact]
-        public void TestPackedRectangles_EmptyInputReturnsNull()
+        public void TestPackedRectangles_EmptyInputReturnsEmptyResult()
         {
             var result = BinPacking.Pack(new List<BinPacking.Rectangle<int>>());
 
-            Assert.Null(result);
+            Assert.NotNull(result);
+            Assert.Empty(result.Rectangles);
+            Assert.Equal(0, result.Area);
+            Assert.Equal(0, result.Width);
+            Assert.Equal(0, result.Height);
+            Assert.Equal(0, result.EmptyPixels);
+            Assert.Equal(double.NaN, result.Filled);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Bearded.Utilities/Algorithms/BinPacking.cs
```diff
@@ -107,11 +108,10 @@ internal Result(ReadOnlyCollection<PositionedRectangle<T>> rectangles, int width
         /// <typeparam name="T">Type of custom user data associated with the rectangle.</typeparam>
         /// <param name="rectangles">The rectangles to pack.</param>
         /// <returns>
-        /// Null, if the given list of rectangles is empty.
-        /// Otherwise, an object containing the packed rectangles and some additional information.
+        /// A result object containing the packed rectangles and some additional information.
         /// </returns>
         /// <exception cref="ArgumentNullException"><paramref name="rectangles"/> is null.</exception>
-        public static Result<T> Pack<T>(IEnumerable<Rectangle<T>> rectangles)
+        public static Result<T>? Pack<T>(IEnumerable<Rectangle<T>> rectangles)
```
tomrijnbeek, line 114: If this algorithm can result null, then I think the documentation should explain what returning null actually means.

Bearded.Utilities/Algorithms/BinPacking.cs
```diff
@@ -123,11 +123,10 @@ public static Result<T> Pack<T>(IEnumerable<Rectangle<T>> rectangles)
         /// If true, the algorithm is run for multiple heuristics and returns the best result.
         /// If false, it is only run for one heuristic, packing rectangles in order of decreasing area, which empirically works well.</param>
         /// <returns>
-        /// Null, if the given list of rectangles is empty.
-        /// Otherwise, an object containing the packed rectangles and some additional information.
+        /// A result object containing the packed rectangles and some additional information.
         /// </returns>
         /// <exception cref="ArgumentNullException"><paramref name="rectangles"/> is null.</exception>
-        public static Result<T> Pack<T>(IEnumerable<Rectangle<T>> rectangles, bool tryMultipleHeuristics)
+        public static Result<T>? Pack<T>(IEnumerable<Rectangle<T>> rectangles, bool tryMultipleHeuristics)
```
tomrijnbeek, line 129: Optional (existing issue):

Booleans in public method interfaces are always a bit of a code smell. The method exerts different behaviours based on the value, so shouldn't it just be two methods? Or is it having more than one responsibility?

Is there a way we can refactor this so that we avoid the boolean? Maybe a specific method for "PackWithMultipleHeuristics", or even a method where you pass in the different heuristics yourself.

Bearded.Utilities/Algorithms/BinPacking.cs
```diff
@@ -173,56 +172,60 @@ private class Tree<T>
...
-                        return W * H;
-                    return down.GetEmptyPixels() + right.GetEmptyPixels();
+                    if (split is var (down, right))
+                        return down.CountEmptyPixels() + right.CountEmptyPixels();
+                    return W * H;
                 }
             }
 
-            private Node root;
+            private Node root = null!;
```
tomrijnbeek, line 228: Optional:

Using `null!` is a bit of a hack IMO. I get why it is allowed here (it is used to share state between different methods called from `Fit`, and it makes sure it's always set), but I had to dig around in the code a bit. Should we adopt the standard where if we use `null!`, we must always add a comment explaining why it's acceptable in this case?

Bearded.Utilities.Tests/Algorithms/BinPackingTests.cs
```diff
@@ -18,7 +18,8 @@ public void TestPackedRectangles_IncludesAll()
 
             var result = BinPacking.Pack(input);
 
-            Assert.Equal(100, result.Rectangles.Count);
+            Assert.NotNull(result);
+            Assert.Equal(100, result!.Rectangles.Count);
```
tomrijnbeek, line 22: Optional:

Have you considered using FluentAssertions if you are rewriting these anyway?

tomrijnbeek (dismissed):
I haven't fully gone into trying to understand the algorithm, but I think the individual changes bring better understanding in general.

Bearded.Utilities/Algorithms/BinPacking.cs
amulware: Right now it's really just a shortcut for not allocating if an empty enumerable is passed in.. Perhaps it would make more sense to return an empty result instead?

Bearded.Utilities/Algorithms/BinPacking.cs
amulware: I'm ok with introducing that rule, but in pretty much all cases we will do this because we're initialising somewhere other than the constructor, but the code still guarantees it will be initialised. Perhaps that in itself shouldn't be done. There may be better options like builder objects or smarter initial values, or...

Bearded.Utilities.Tests/Algorithms/BinPackingTests.cs
amulware: Happy to do so if you like, I wanted to keep the changes to a minimum.

Bearded.Utilities/Algorithms/BinPacking.cs
tomrijnbeek: Agreed

Bearded.Utilities/Algorithms/BinPacking.cs
tomrijnbeek: +1 that avoiding this situation is better than this hack. Forcing you to comment would force you to think about it too in a way.

Bearded.Utilities.Tests/Algorithms/BinPackingTests.cs
tomrijnbeek: We can follow up separately.

Bearded.Utilities/Algorithms/BinPacking.cs
amulware: There are now separate methods.

Bearded.Utilities/Algorithms/BinPacking.cs
amulware: Found a way to avoid it, which is nice.

amulware:
Implemented all changes as requested and made other improvements. I hope you don't mind me splitting it out into multiple files, I think it's a lot more readable this way.

Note that:
- I changed one test, since the result now will never be null.
- We are only testing one of the now FOUR pack methods. That doesn't seem right of course - do you agree I should test all four? If so, I'd propose I do this as part of a rewrite of the tests using an abstract class, so I don't have to repeat the exact same test code (minus the method call) four times. Your thoughts?

I'd be happy merging it like this and creating a follow up issue for the tests. Alternatively, I could fix tests in this PR as well, if you prefer.

tomrijnbeek (approved):
Separate PR for testing changes sounds good.

PR #195, opened 2020-10-12
0️⃣ Start enabling nullable

### LINKED ISSUE #184: Enforce nullable reference types

amulware:
## ✨ What's this?
This enables nullable as errors in all projects and makes the necessary changes.

### 🔗 Relationships

Ref #184
Ref #194 (a pre-requisite for this PR)

## 🔍 Why do we want this?
Nullability is hip.

## 🏗 How is it done?
I enabled nullable and made sure they end up as errors in all project files. Then it was as simple as hunting errors and fixing case by case until everything compiled again.

### 💥 Breaking changes
This is a breaking change for anyone who has nullable enabled themselves, as some of our public members are now explicitly nullable.

### 🔬 Why not another way?
This is the most straight forward. The changes are pretty minor and splitting it into multiple PRs seemed silly, with the exception of #194, which includes several major changes.

## 💡 Review hints
I'll leave comments on all changes to explain why they are correct. Feel free to review based on that or however else you like.

Note that until #194 is merged into master and this PR brought up to date, the main project will not build. There is no conflict between the changes however.

### DIFF
--- a/Bearded.Utilities.Testing.Tests/Bearded.Utilities.Testing.Tests.csproj
+++ b/Bearded.Utilities.Testing.Tests/Bearded.Utilities.Testing.Tests.csproj
@@ -2,9 +2,11 @@
   <PropertyGroup>
     <TargetFrameworks>net472;netcoreapp2.2</TargetFrameworks>
     <LangVersion>8</LangVersion>
-    <Nullable>warnings</Nullable>
+    <Nullable>enable</Nullable>
+    <WarningsAsErrors>CS8600;CS8602;CS8603</WarningsAsErrors>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.abstractions" Version="2.0.3" />
     <PackageReference Include="xunit.runner.console" Version="2.4.1" />
--- a/Bearded.Utilities.Testing.Tests/Bearded.Utilities.Testing.Tests.csproj
+++ b/Bearded.Utilities.Testing.Tests/Bearded.Utilities.Testing.Tests.csproj
@@ -3,7 +3,7 @@
     <TargetFrameworks>net472;netcoreapp2.2</TargetFrameworks>
     <LangVersion>8</LangVersion>
     <Nullable>enable</Nullable>
-    <WarningsAsErrors>nullable</WarningsAsErrors>
+    <WarningsAsErrors>CS8600;CS8602;CS8603</WarningsAsErrors>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.1" />
--- a/Bearded.Utilities.Testing.Tests/Bearded.Utilities.Testing.Tests.csproj
+++ b/Bearded.Utilities.Testing.Tests/Bearded.Utilities.Testing.Tests.csproj
@@ -3,7 +3,7 @@
     <TargetFrameworks>net472;netcoreapp2.2</TargetFrameworks>
     <LangVersion>8</LangVersion>
     <Nullable>enable</Nullable>
-    <WarningsAsErrors>CS8600;CS8602;CS8603</WarningsAsErrors>
+    <WarningsAsErrors>nullable</WarningsAsErrors>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.1" />
--- a/Bearded.Utilities/Core/Maybe.cs
+++ b/Bearded.Utilities/Core/Maybe.cs
@@ -67,7 +67,7 @@ namespace Bearded.Utilities
 
     public static class Maybe
     {
-        public static Maybe<T> FromNullable<T>(T value) where T : class =>
+        public static Maybe<T> FromNullable<T>(T? value) where T : class =>
             value == null ? Nothing : Maybe<T>.Just(value);
 
         public static Maybe<T> FromNullable<T>(T? value) where T : struct =>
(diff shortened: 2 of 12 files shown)


Bearded.Utilities/Core/Maybe.cs
```diff
@@ -67,7 +67,7 @@ public TResult Match<TResult>(Func<T, TResult> onValue, Func<TResult> onNothing)
 
     public static class Maybe
     {
-        public static Maybe<T> FromNullable<T>(T value) where T : class =>
+        public static Maybe<T> FromNullable<T>(T? value) where T : class =>
```
amulware, line 70: Without this, we can't create Nothing from a reference type.

Bearded.Utilities/IO/Logger.cs
```diff
@@ -30,20 +30,20 @@ internal Writer(Logger logger, Severity severity)
             }
 
             public void Log<T>(T obj) {
-                logger.AddEntry(new Entry(obj.ToString(), severity));
+                logger.AddEntry(new Entry(obj?.ToString() ?? "null", severity));
```
amulware, line 33: This actually looks like it was a bug - it would have been trivial to blow up the logger by passing in `null`. No longer. Though I am not 100% sure if outputting 'null' is the best fallback in this case. Generally, `null` prints as empty string in string formatting, but I felt that an empty string, or ignoring the `Log()` call entirely would make it hard to figure out that null was logged. Being explicit about it will make it easier for the user to find errors in their code.

Bearded.Utilities/IO/Logger.cs
```diff
@@ -30,20 +30,20 @@ internal Writer(Logger logger, Severity severity)
...
-        public Writer Warning { get; }
-        public Writer Info { get; }
-        public Writer Debug { get; }
-        public Writer Trace { get; }
+        public Writer? Fatal { get; }
+        public Writer? Error { get; }
+        public Writer? Warning { get; }
+        public Writer? Info { get; }
+        public Writer? Debug { get; }
+        public Writer? Trace { get; }
```
amulware, line 46: This is the only real breaking change of this PR. We always intended the logger to be used like `logger.Warning?.Log()` so this is finally enforcing that to be best of our ability.

Bearded.Utilities/IO/Logger.cs
```diff
@@ -272,17 +272,17 @@ private static void ensureListCapacity(List<Entry> list, int neededCapacity)
...
+            Warning = createWriterIfSeverityEnabled(enabledSeverities, Severity.Warning);
+            Info = createWriterIfSeverityEnabled(enabledSeverities, Severity.Info);
+            Debug = createWriterIfSeverityEnabled(enabledSeverities, Severity.Debug);
+            Trace = createWriterIfSeverityEnabled(enabledSeverities, Severity.Trace);
 
             RecentEntries = lines.AsReadOnly();
         }
 
-        private Writer getWriterIfSeverityEnabled(ICollection<Severity> enabledSeverities, Severity severity)
+        private Writer? createWriterIfSeverityEnabled(ICollection<Severity> enabledSeverities, Severity severity)
```
amulware, line 285: The ? is obvious, but I also renamed it because it creates, it does not 'get' a writer.

Bearded.Utilities/Linq/Extensions.cs
```diff
@@ -234,7 +234,7 @@ public static T RandomElement<T>(this IEnumerable<T> source, Random random)
             {
                 throw new InvalidOperationException("Sequence was empty.");
             }
-            return current;
+            return current!;
```
amulware, line 237: The above check on `count` ensures that this will never be null (unless in fact the sequence contained a null value, in which case that is the intended behaviour).

Bearded.Utilities/Threading/BackgroundActionQueue.cs
```diff
@@ -129,7 +129,7 @@ public T RunAndReturn<T>(Func<T> action)
         {
             var ret = default(T);
             RunAndAwait(() => ret = action());
-            return ret;
+            return ret!;
```
amulware, line 132: The implementation of `RunAndAwait` ensures that the return value will be set. It may of course be set to null, if that is what the action returns, but that is entirely up to the caller.

Bearded.Utilities/IO/Logger.cs
tomrijnbeek: I am okay with using a null string here.

tomrijnbeek (changes requested):
I think all my comments are optional, but it would be good to try if we can replace the warning numbers with the "nullable" string as suggested.

Bearded.Utilities.Testing.Tests/Bearded.Utilities.Testing.Tests.csproj
```diff
@@ -2,9 +2,11 @@
   <PropertyGroup>
     <TargetFrameworks>net472;netcoreapp2.2</TargetFrameworks>
     <LangVersion>8</LangVersion>
-    <Nullable>warnings</Nullable>
+    <Nullable>enable</Nullable>
+    <WarningsAsErrors>CS8600;CS8602;CS8603</WarningsAsErrors>
```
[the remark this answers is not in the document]
amulware: This is great, thanks! Turns out this causes a ton of other errors, so I will leave it at the above warnings for now and submit further PRs to achieve full nullability for easier review.

amulware:
Explaining the null forgiving operators made me find another issue: https://github.com/beardgame/utilities/issues/197 so that's great.

I'd like to merge it like it is now (after the bin packing PR), and then submit another PR with all nullable errors fixed (or perhaps multiple if there are other things that require big changes).

tomrijnbeek (approved):

PR #202, opened 2020-10-17
🧓 Deprecate Mathf

### LINKED ISSUE #201: Deprecate Mathf in favour of System MathF
## 🆙 Improvement proposal
No need to implement our own Mathf functions anymore, as System libraries now contain their own MathF class.

## Context
It can be reasonably expected that System classes will be as performant as our own, if not more.

tomrijnbeek:
## ✨ What's this?
This PR deprecates the `Mathf` class in favour of other classes.

### 🔗 Relationships
Ref #201 

## 🔍 Why do we want this?
.NET Core includes a `System.MathF` class which contains a copy of all math functions for floats. `Mathf` was originally introduced with this purpose, but is now obsolete. This is why this PR deprecates the class in favour of the standard System one, which is much more complete.

## 🏗 How is it done?
All existing functions on `Mathf`, including the class itself, is marked as obsolete to give time to migrate to the new functionality.

Not all functionality on `Mathf` actually mirrors `Math`. In my opinion, this means it had already grown beyond its original purpose. The remaining functionality is split out in `MathConstants`, providing a much tighter responsibility, and `MoreMath`, containing functionality that I couldn't place anywhere else.

### 💥 Breaking changes
No breaking changes at this time, unless people treat obsolete warnings as errors.

### 🔬 Why not another way?
I believe this is the correct separation of concerns.

### 🦋 Side effects
`Mathf` will have to be truly and really deleted in a future PR.

## 💡 Review hints
None of the code should be new.

### DIFF
--- a/Bearded.Utilities/Core/MathConstants.cs
+++ b/Bearded.Utilities/Core/MathConstants.cs
@@ -1,0 +1,42 @@
+using System;
+
+namespace Bearded.Utilities
+{
+    public static class MathConstants
+    {
+        /// <summary>
+        /// Represents the square root of 2 (1.414213f).
+        /// </summary>
+        public const float Sqrt2 = 1.414213f;
+
+        /// <summary>
+        /// Represents the square root of 3 (1.732051f).
+        /// </summary>
+        public const float Sqrt3 = 1.732051f;
+
+        /// <summary>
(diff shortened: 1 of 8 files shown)


amulware (approved):
Do please create a follow up issue or not close the current one and leave a note.

PR #203, opened 2020-10-18
👷‍♂️ Add GitHub actions work flow for building/testing and releasing

tomrijnbeek:
This PR adds a GitHub actions config.

It's based on the default template as provided by GitHub, from which I then built a release flow as well.

This flow will automatically build and test on every PR and master branch bunch (see the example in this PR).

When a tag is pushed to the default branch, it will:

* Execute the normal "build" job as if it is a normal push (this generates the necessary binary files, and also makes sure the tests are still passing)
* Package the nupkg files
* Upload the nupkg files to NuGet using an API token. The API token has to be stored as a secret in the organisation.
* Upload the nupkg files to GitHub packages. This uses the default GitHub secret token.
* Create a GitHub release. We currently don't attach any custom assets to this release, since there isn't a way to fetch all the nupkg files using a file glob.

### DIFF
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -1,0 +1,25 @@
+name: .NET Core
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup .NET Core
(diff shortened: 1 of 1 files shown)


amulware:
I don't know what's going on here, but it shows warnings straight in the changes. This is very cool.

amulware (approved):
I won't pretend I had the time to get into this and understand all of it, but the broad strokes make sense, and as discussed elsewhere, I'm happy to give it a go and see what happens!

Issue #198, opened 2020-10-14
Migrate to GitHub actions


tomrijnbeek:
## 🆙 Improvement proposal
Currently we use AppVeyor. GitHub actions now has native CI, which may be better for us.

## Context
GitHub actions are directly integrated into GitHub, so this may give us better support. On top of that, GitHub actions configs can be reused between repositories.

PR #210, opened 2020-10-20
🐛 Fix two additional GitHub actions problems

tomrijnbeek:
## ✨ What's this?
Another two bug fixes:

* Make sure we build the release configuration, not the debug one.
* Fix the path to detect NuGet packages.

### 🔗 Relationships
See #198


amulware (approved):
✨

PR #218, opened 2021-01-28
Improve casts in Equals(object) (#204)

### LINKED ISSUE #204: Improve casts in Equals(object)
Many of our types implement Equals(object) like:
```cs
public override bool Equals(object? obj) => obj is Rectangle && Equals((Rectangle) obj);
```

We can improve these to use one less cast like:
```cs
public override bool Equals(object? obj) => obj is Rectangle rectangle && Equals(rectangle);
```

We should do so with all cases for improvements to performance (and maybe readability).

Jlll1:
## ✨ What's this?
As described in the issue, it reduces the amount of casts in Equals() comparations.

### 🔗 Relationships

Closes #204

### DIFF
--- a/Bearded.Utilities/Geometry/Angle.cs
+++ b/Bearded.Utilities/Geometry/Angle.cs
@@ -290,7 +290,7 @@ namespace Bearded.Utilities.Geometry
         /// </returns>
         public override bool Equals(object? obj)
         {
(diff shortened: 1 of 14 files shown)


tomrijnbeek (approved):
This is fantastic, thanks for the contribution!

PR #219, opened 2021-01-29
🔠 Add format-only ToString to SpaceTime types

### LINKED ISSUE #196: Review formatting ToString implementations
## 🆙 Improvement proposal
All spacetime types (and perhaps others) have ToString implementations which take a format and format provider. We should consider adding another overload that only requires a format, and make that nullable, to mirror .Net's numeric type ToString overloads.

Jlll1:
## ✨ What's this?
This adds another overload for ToString methods, that takes nullable format as its only parameter. I've also normalized the implementations to follow similar formatting.

### 🔗 Relationships

Closes #196

### DIFF
--- a/Bearded.Utilities/Geometry/Angle.cs
+++ b/Bearded.Utilities/Geometry/Angle.cs
@@ -358,6 +358,9 @@ namespace Bearded.Utilities.Geometry
         public string ToString(string? format, IFormatProvider? formatProvider)
             => $"{Degrees.ToString(format, formatProvider)}°";
 
+        public string ToString(string? format)
+            => $"{Degrees.ToString(format)}°";
+
         #endregion
 
         #region Casts
--- a/Bearded.Utilities/Geometry/Angle.cs
+++ b/Bearded.Utilities/Geometry/Angle.cs
@@ -359,7 +359,7 @@ namespace Bearded.Utilities.Geometry
             => $"{Degrees.ToString(format, formatProvider)}°";
 
         public string ToString(string? format)
-            => $"{Degrees.ToString(format)}°";
+            => ToString(format, CultureInfo.CurrentCulture);
 
         #endregion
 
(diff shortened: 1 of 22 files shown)


Bearded.Utilities/Geometry/Angle.cs
```diff
@@ -358,6 +358,9 @@ public override int GetHashCode()
         public string ToString(string? format, IFormatProvider? formatProvider)
             => $"{Degrees.ToString(format, formatProvider)}°";
 
+        public string ToString(string? format)
+            => $"{Degrees.ToString(format)}°";
```
amulware, line 362: This applies for all these new methods: Since this uses the exact same format string as the version with IFormatProvider, it would be great if the new method would call that other one instead, either passing in null or CultureInfo.CurrentCulture (I'm not sure which is preferred, but I think they do basically the same).

Bearded.Utilities/SpaceTime/2d/Acceleration2.cs
```diff
@@ -131,8 +131,13 @@ public Acceleration2(Acceleration x, Acceleration y)
 
         public override string ToString() => ToString(null, CultureInfo.CurrentCulture);
 
-        public string ToString(string? format, IFormatProvider? formatProvider)
-            => $"({value.X.ToString(format, formatProvider)}, {value.Y.ToString(format, formatProvider)}) u/t²";
+        public string ToString(string? format, IFormatProvider? formatProvider) => "(" +
+            $"{value.X.ToString(format, formatProvider)}, " +
+            $"{value.Y.ToString(format, formatProvider)}) u/t²";
```
amulware, line 136: Here and some others below: I would prefer if we didn't change the existing implementations unless we actually intend their behaviour to change (or if there is a really good reason - is there?), so that the change set is easier to understand.

amulware (changes requested):
Hey, thanks for the PR!

I can't see anything functionally wrong with that, but left some thoughts in comments.

Bearded.Utilities/Geometry/Angle.cs
Jlll1: I was actually thinking to do it that way, but for some mystical reason I went with this one, I'll get right on it.

amulware (approved):

amulware:
Thanks for the quick changes!

PR #220, opened 2021-02-16
👷‍♂️ Add GitHub actions automation

tomrijnbeek:
## ✨ What's this?
Replaces Appveyor with GitHub Actions for CI.

## 🔍 Why do we want this?
Can more easily reuse templates across our repositories this way.

## 🏗 How is it done?
Copy the templates from the beardgame/.github repository.

## 💡 Review hints
1. If there are problems with the actual GitHub actions, open an issue/PR on beardgame/.github instead
1. Appveyor is failing due to deleted Appveyor file. This is fine, I will turn off Appveyor once this PR is approved.


amulware (approved):
Did not test, but looks good at a glance.

PR #226, opened 2021-06-02
✅ Add priority queue enumeration test

amulware:
## ✨ What's this?
Adds a test for priority queue enumeration.

### 🔗 Relationships
Follow up to #225

## 🔍 Why do we want this?
Tests.

## 🏗 How is it done?
A single pretty big tests that queues and dequeues some items and then checks that the enumeration:
- has the correct count
- are all items we added before
- are items that are actually in the queue

### 🔬 Why not another way?
Assuming Enqueue/Dequeue work correctly, this checks a lot of potential issues that could happen, especially with it being a property test that takes a random seed.

However, it's also a bit of a complex test. If you'd like I can simplify it or replace it with multiple.

### DIFF
--- a/Bearded.Utilities.Tests/Collections/PriorityQueueTests.cs
+++ b/Bearded.Utilities.Tests/Collections/PriorityQueueTests.cs
@@ -2,6 +2,9 @@
 using System.Collections.Generic;
 using System.Linq;
 using Bearded.Utilities.Collections;
+using Bearded.Utilities.Linq;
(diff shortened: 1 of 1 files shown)


tomrijnbeek (approved):
Yeah, I would generally try and extract different behaviours into different tests, i.e. test one behaviour per test, not one method. I think this is probably the best way to ensure that you get exactly the elements in the queue, and nothing more.

PR #227, opened 2021-06-08
🧵 Make IActionQueue async

amulware:
## ✨ What's this?
Refactored the IActionQueue interface and implementors to be simpler and properly Async, making them easier to use in async cases (at the cost of a little bit of overhead of creating task completion sources).

## 🔍 Why do we want this?
Working with Tasks and async/await is nicer than doing stuff more manually.

## 🏗 How is it done?
See code. Quite simple.

### 💥 Breaking changes
The whole interface is different now.

### 🔬 Why not another way?
This is the simplest I think.

### DIFF
--- a/Bearded.Utilities/Threading/ManualActionQueue.cs
+++ b/Bearded.Utilities/Threading/ManualActionQueue.cs
@@ -1,12 +1,12 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Diagnostics;
-using System.Threading;
+using System.Threading.Tasks;
 
 namespace Bearded.Utilities.Threading
 {
     /// <summary>
-    /// A threadsafe queue to run actions from.
+    /// A thread-safe queue to run actions from.
     /// Typical usage is to schedule actions from multiple threads and execute them on one main thread.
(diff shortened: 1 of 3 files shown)


Bearded.Utilities/Threading/ManualActionQueue.cs
```diff
@@ -85,96 +81,35 @@ public int ExecuteFor(TimeSpan time)
...
 
-        /// <summary>
-        /// Queues an action to run. Returns only after the action has been executed.
-        /// </summary>
-        /// <param name="action">The action to run.</param>
-        public void RunAndAwait(Action action)
+        public Task Run(Action action)
         {
-            var reset = new ManualResetEvent(false);
+            var task = new TaskCompletionSource<object?>();
```
tomrijnbeek, line 91: nit: should this be Void?

tomrijnbeek (approved):

Bearded.Utilities/Threading/ManualActionQueue.cs
amulware: From what I've seen, using object and null is somewhat of a convention.

PR #228, opened 2021-07-28
⭕ Implement circular arcs

tomrijnbeek:
## ✨ What's this?
This PR adds a new geometry type for circular arcs in 2D space.

## 🔍 Why do we want this?
Circular arcs are a pretty useful geometric shape, so having a way to represent them first of all is quite helpful. Additionally, I plan to send follow-up PRs that allow simple geometric constructions such as constructing a circular arc given two lines and a radius (to automatically create rounded corners).

## 🏗 How is it done?
The `CircularArc2` is a new readonly struct keeping track of a `center`, `radius`, `start` and `angle`. It provides several properties to easily construct similar arcs, and there are several factory functions to create arcs in different ways.

### 💥 Breaking changes
N/A

### 🔬 Why not another way?
There are some other representations that I considered. The most obvious alternative is to keep start and end, instead of start and angle. However, two points on a circle define both the major and minor arc, so we would have to keep track of a separate bit. Using the angle to store that bit makes sense. Other representations are less convenient to calculate back from (e.g. circle center and two points), since most interesting properties of an arc follow almost trivially from the stored attributes, while restoring these attributes from those properties is generally non-trivial.

### 🦋 Side effects
Future PRs will provide further functions to construct these arcs when the attributes aren't known yet.

## 💡 Review hints
The added tests were purposefully not meant to provide 100% coverage, but should at least cover all the interesting construction logic.

### DIFF
--- a/Bearded.Utilities/Geometry/CircularArc2.cs
+++ b/Bearded.Utilities/Geometry/CircularArc2.cs
@@ -1,0 +1,89 @@
+using System;
+using OpenTK.Mathematics;
+
+namespace Bearded.Utilities.Geometry
+{
+    /// <summary>
+    /// Represents an arc of a circle between a pair of distinct points.
+    /// </summary>
+    public readonly struct CircularArc2 : IEquatable<CircularArc2>
+    {
+        public Vector2 Center { get; }
+        public float Radius { get; }
+        public Direction2 Start { get; }
+        public Angle Angle { get; }
+
+        public Direction2 End => Start + Angle;
+        public Vector2 StartPoint => Center + Radius * Start.Vector;
+        public Vector2 EndPoint => Center + Radius * End.Vector;
+
+        public float ArcLength => Angle.Radians * Radius;
+
+        /// <summary>
+        /// Returns the major arc if this arc is the minor arc, and returns the major arc if this is the minor arc.
+        /// </summary>
+        public CircularArc2 Opposite =>
+            new CircularArc2(Center, Radius, Start, Angle.Sign() * -(MathConstants.TwoPi.Radians() - Angle.Abs()));
+
+        /// <summary>
+        /// Returns the same arc with the start and end directions swapped.
+        /// </summary>
+        public CircularArc2 Reversed => new CircularArc2(Center, Radius, End, -Angle);
+
+        /// <summary>
+        /// Constructs the minor arc in the circle with specified center and radius between the given directions.
+        /// </summary>
+        public static CircularArc2 ShortestArcBetweenDirections(
+            Vector2 center, float radius, Direction2 start, Direction2 end)
+        {
+            return new CircularArc2(center, radius, start, end - start);
+        }
+
+        /// <summary>
+        /// Constructs the major arc in the circle with specified center and radius between the given directions.
+        /// </summary>
+        public static CircularArc2 LongestArcBetweenDirections(
+            Vector2 center, float radius, Direction2 start, Direction2 end)
+        {
+            var shortestAngle = end - start;
+            var longestAngle = -(MathConstants.TwoPi.Radians() - shortestAngle);
+
+            return new CircularArc2(center, radius, start, longestAngle);
+        }
+
+        /// <summary>
+        /// Constructs the arc in the circle with specified center and radius starting in the given direction,
+        /// subtending the given angle.
+        /// </summary>
+        public static CircularArc2 FromStartAndAngle(
+            Vector2 center, float radius, Direction2 start, Angle angle)
+        {
+            return new CircularArc2(center, radius, start, angle);
+        }
+
+        private CircularArc2(Vector2 center, float radius, Direction2 start, Angle angle)
+        {
+            Center = center;
+            Radius = radius;
+            Start = start;
+            Angle = angle;
+        }
+
+        public bool Equals(CircularArc2 other) =>
+            Center.Equals(other.Center)
+            && Radius.Equals(other.Radius)
+            && Start.Equals(other.Start)
+            && Angle.Equals(other.Angle);
+
+        public override bool Equals(object? obj) => obj is CircularArc2 other && Equals(other);
+
+        public override int GetHashCode() => HashCode.Combine(Center, Radius, Start, Angle);
+
+        public static bool operator ==(CircularArc2 left, CircularArc2 right) => left.Equals(right);
+        public static bool operator !=(CircularArc2 left, CircularArc2 right) => !left.Equals(right);
+
+        public override string ToString() =>
+            string.Join(System.Environment.NewLine,
+                $"Center = {Center}", $"Radius = {Radius}", $"Start = {Start}", $"Angle = {Angle}");
+    }
+}
--- a/Bearded.Utilities.Tests/Geometry/CircularArc2Tests.cs
+++ b/Bearded.Utilities.Tests/Geometry/CircularArc2Tests.cs
@@ -1,0 +1,98 @@
+using Bearded.Utilities.Geometry;
+using FluentAssertions;
+using FluentAssertions.Equivalency;
+using OpenTK.Mathematics;
+using Xunit;
+
+namespace Bearded.Utilities.Tests.Geometry
+{
+    public sealed class CircularArc2Tests
+    {
+        public sealed class Opposite
+        {
+            [Fact]
+            public void ReturnsMinorArcIfArcIsMajor()
+            {
+                var arc = CircularArc2.FromStartAndAngle(Vector2.Zero, 1, Direction2.Zero, -270.Degrees());
+
+                var actual = arc.Opposite;
+
+                var expected = CircularArc2.FromStartAndAngle(Vector2.Zero, 1, Direction2.Zero, 90.Degrees());
+                actual.Should().BeEquivalentTo(expected, circularArc2Equivalency);
+            }
+
+            [Fact]
+            public void ReturnsMajorArcIfArcIsMinor()
+            {
+                var arc = CircularArc2.FromStartAndAngle(Vector2.Zero, 1, Direction2.Zero, 90.Degrees());
+
+                var actual = arc.Opposite;
+
+                var expected = CircularArc2.FromStartAndAngle(Vector2.Zero, 1, Direction2.Zero, -270.Degrees());
+                actual.Should().BeEquivalentTo(expected, circularArc2Equivalency);
+            }
+        }
+
+        public sealed class Reversed
+        {
+            [Fact]
+            public void ReturnsReversedArc()
+            {
+                var arc = CircularArc2.FromStartAndAngle(Vector2.Zero, 1, Direction2.Zero, 90.Degrees());
+
+                var actual = arc.Reversed;
+
+                var expected =
+                    CircularArc2.FromStartAndAngle(Vector2.Zero, 1, Direction2.FromDegrees(90), -90.Degrees());
+                actual.Should().BeEquivalentTo(expected, circularArc2Equivalency);
+            }
+        }
+
+        public sealed class ShortestArcBetweenDirections
+        {
+            [Fact]
+            public void ReturnsMinorArc()
+            {
+                var actual = CircularArc2.ShortestArcBetweenDirections(
+                    Vector2.Zero,
+                    1,
+                    Direction2.FromDegrees(0),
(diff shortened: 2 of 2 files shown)


Bearded.Utilities/Geometry/CircularArc2.cs
```diff
@@ -0,0 +1,95 @@
...
+        public static bool operator ==(CircularArc2 left, CircularArc2 right) => left.Equals(right);
+        public static bool operator !=(CircularArc2 left, CircularArc2 right) => !left.Equals(right);
+
+        public override string ToString() =>
+            string.Join(System.Environment.NewLine,
+                $"Center = {Center}", $"Radius = {Radius}", $"Start = {Start}", $"Angle = {Angle}");
+
+        private static Angle oppositeAngle(Angle original)
+        {
+            return -original.Sign() * (MathConstants.TwoPi.Radians() - original.Abs());
```
amulware, line 92: This math is difficult for me to follow - can we use a local variable or some other way of making it more explicit what it is doing?

Bearded.Utilities/Geometry/CircularArc2.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// Returns the same arc with the start and end directions swapped.
+        /// </summary>
+        public CircularArc2 Reversed => new CircularArc2(Center, Radius, End, -Angle);
+
+        /// <summary>
+        /// Constructs the minor arc in the circle with specified center and radius between the given directions.
+        /// </summary>
+        public static CircularArc2 ShortestArcBetweenDirections(
```
amulware, line 35: This name makes me feel like maybe there's more than one possible Arc here. Would `ShorterArcBetweenDirections` be more correct? Or maybe even just `ShortArc..` or `MinorArc..` to use the mathematical term?

Bearded.Utilities/Geometry/CircularArc2.cs
```diff
@@ -0,0 +1,95 @@
...
+        public override bool Equals(object? obj) => obj is CircularArc2 other && Equals(other);
+
+        public override int GetHashCode() => HashCode.Combine(Center, Radius, Start, Angle);
+
+        public static bool operator ==(CircularArc2 left, CircularArc2 right) => left.Equals(right);
+        public static bool operator !=(CircularArc2 left, CircularArc2 right) => !left.Equals(right);
+
+        public override string ToString() =>
+            string.Join(System.Environment.NewLine,
+                $"Center = {Center}", $"Radius = {Radius}", $"Start = {Start}", $"Angle = {Angle}");
```
amulware, line 88: Should relatively primitive types like this serialize to multi-line strings?

amulware (changes requested):
The tests all hardcode direction and angle to multiples of 90 degrees. Could we turn them into property tests to make sure less trivial cases are covered too? Or is there an argument against that?

amulware:
The one argument I could see for storing two directions and a bool for minor/major (or a sign) would be that the binary representation of Direction is very neat and can work with great accuracy and predictability. That being said, the current representation doesn't bother me much and might be perfectly fine.

tomrijnbeek:
> The one argument I could see for storing two directions and a bool for minor/major (or a sign) would be that the binary representation of Direction is very neat and can work with great accuracy and predictability. That being said, the current representation doesn't bother me much and might be perfectly fine.

After thinking about it a bit more, I think you're right. While it may be awkward to store a boolean separately, it means that the math becomes much nicer, and there's no weird situation where doing the same operation that is its own inverse leads to a result that is slightly different than what you started with due to float imprecision.

amulware:
I'm not sure if we can exactly count on that, given that you'd probably do those operations using floating point angles... but yeah, if the code turns out well, go with it.

Bearded.Utilities/Geometry/CircularArc2.cs
tomrijnbeek: I understand your point. I've renamed it to `MinorArc` now, but I want to explain my hesitation at not doing so originally. While minor and major arc are indeed the correct mathematical terms here, I think that most programmers to which this library would be aimed do not know. Even I wasn't aware that these were the terms until I ran into them. `Shortest` and `Longest` made more sense to me. `Shorter`... eh, I guess it works. `Short`... perhaps. LMK if you think using the mathematical term is a concern to you, and which alternative you prefer in that case.

amulware: I'm happy using the mathematical term, however, I also had to look it up to be sure which is which. So... I'm not sure how helpful that is. Of all the non-mathematical ones 'Short' seems the best because it is concise, relatively ambiguous, and makes for good prose "get me the short arc between these directions".

Btw, does Direction2 have two ways of creating an angle from two directions and if so what are they called?

Bearded.Utilities/Geometry/CircularArc2.cs
tomrijnbeek: I could not think of a succinct notation that is still easy to interpret. There is no official notation for a circle with center `P` and radius `r` as far as I can tell, so representing the circle is the first thing I'm stuck at. Arcs are also generally notated using the coordinates of the three points, which doesn't map to our representation correctly.

The best thing I can think of is `(x, y, z) × r; minor/major arc α rad - β rad`

Bearded.Utilities/Geometry/CircularArc2.cs
tomrijnbeek: `Short` and `Long` it is then.

It doesn't seem like `Angle` has a way to get the long angle between directions, it always return the shortest one. This is called `Angle.Between`. There are also `BetweenPositive` and `BetweenNegative` which always return a positive and negative angle respectively.

Bearded.Utilities/Geometry/CircularArc2.cs
tomrijnbeek: I updated it to use variables. I think this can also be solved using branching code, but I think this still has my preference.

amulware: Yeah, now it's super clear, I'm happy with that!

tomrijnbeek:
> The tests all hardcode direction and angle to multiples of 90 degrees. Could we turn them into property tests to make sure less trivial cases are covered too? Or is there an argument against that?

My primary concern is that to do the construction tests that were there using randomly generated directions would have meant pretty much copying the logic into the test. The whole reason I tested the construction is that I didn't trust myself to implement it correctly, so why would I rely on doing it correctly in the tests?

I don't disagree that property tests could help here, but I think they should test what they say they test: properties. A short arc should always have an angle less than pi for example. However, even if I had tested those properties, I'm still not sure if I would have had the correct implementation.

Now the construction is much more straightforward. I'm happy to add property tests in addition to the existing tests, though I don't know if they would fully replace them.

amulware (commented):
It looks good, but there's one edge case I'm unsure about: If Start == End, then the Sign() of the angle is 0, and the opposite magnitude will be 0, hence the opposite arc will be the same as the original.

I'm not sure how to correctly handle this case though, since there are sort of two major arcs, one in each direction, and just picking one of them seems wrong.

Thoughts? Does this edge case have a mathematically defined solution?

tomrijnbeek:
> It looks good, but there's one edge case I'm unsure about: If Start == End, then the Sign() of the angle is 0, and the opposite magnitude will be 0, hence the opposite arc will be the same as the original.
> 
> I'm not sure how to correctly handle this case though, since there are sort of two major arcs, one in each direction, and just picking one of them seems wrong.
> 
> Thoughts? Does this edge case have a mathematically defined solution?

Yeah, that's a good point. We could always say that in this case the minor arc (which is just a point) is in the positive direction (counter-clockwise in your standard x right y up coordinate system), and the major arc (which is the full circle) is in the negative direction, but that's very arbitrary. The alternative is we don't allow it, but then suddenly the default instance of this struct is invalid, which I don't like either.

I don't have a great solution...

tomrijnbeek:
Thinking about it more, I think we can adopt the following:

*If `(end - start) == (start - end)`, then the short arc is defined to be the arc with the positive angle.*

This can happen in exactly two situations.

1. `start == end`, in which case the short arc is defined to be the point, and the long arc is defined to be the whole circle with negative directionality. Note that this explicitly means that a full 360° arc traversed positively can never be represented by this type. I do not believe this is a concern (you should probably be using a different type instead).
2. `start == end + 180°`, in which case the two arcs are the same length, and the arc traversing in the positive direction is considered the shorter one.

To achieve and maintain this, the changes that need to be made to the code, I believe, are:

* Ensure the result of `Direction2 - Direction2` indeed returns 180° and not -180° in case the directions are direct opposites.
* Forbid `CircularArc2.FromDirectionAndAngle` to be called with an angle larger than **or equal** 2π rather than just larger than.
* Add special logic in `oppositeAngle` to handle the case where `Angle.Sign()` returns 0.

amulware:
Good point about the 180-degrees case. Your proposal sounds as good as any... better to have decided on _some_ behaviour than keep it undefined. (Though I'm happy to go with whatever Direction2 - Direction2 does right now honestly :D)

Let's have tests for these edge cases though to make sure we don't break things.

Bearded.Utilities.Tests/Generators/DirectionGenerators.cs
```diff
@@ -0,0 +1,14 @@
...
+using FsCheck;
+
+namespace Bearded.Utilities.Tests.Generators
+{
+    static class DirectionGenerators
+    {
+        public static class All
+        {
+            public static Arbitrary<Direction2> Directions() =>
+                Arb.Generate<uint>().Select(i => Direction2.FromDegrees(360f / (1L << 32))).ToArbitrary();
```
amulware, line 11: This makes me realise that it may be impossible to easily construct many possible Direction2 values since floats have less significant digits than the underlying integer. Curious.

Bearded.Utilities/Geometry/CircularArc2.cs
```diff
@@ -0,0 +1,126 @@
...
+                throw new ArgumentException(
+                    "Angle must be between -360° inclusive and +360° exclusive.", nameof(angle));
+            }
+
+            // When the start and end are opposite each other, the short arc goes in the negative direction (i.e. the
+            // angle is -π), so conversely -π should be considered a short arc. The long arc goes in the positive
+            // direction (i.e. the angle is -π), so conversely -π should be considered a long arc.
+            var isShortArc = angle.Radians >= -MathConstants.Pi && angle.Radians < MathConstants.Pi;
+            return new CircularArc2(center, radius, start, start + angle, isShortArc);
+        }
```
amulware, line 83:
```suggestion
        {
            var radians = angle.Radians;
            if (radians < -MathConstants.TwoPi || radians >= MathConstants.TwoPi)
            {
                throw new ArgumentException(
                    "Angle must be between -360° inclusive and +360° exclusive.", nameof(angle));
            }

            // When the start and end are opposite each other, the short arc goes in the negative direction (i.e. the
            // angle is -π), so conversely -π should be considered a short arc. The long arc goes in the positive
            // direction (i.e. the angle is -π), so conversely -π should be considered a long arc.
            var isShortArc = radians >= -MathConstants.Pi && radians < MathConstants.Pi;
            return new CircularArc2(center, radius, start, start + angle, isShortArc);
        }
```

The repetition bothered me a bit, even though I know the implementation. Completely optional, if you think this makes it less readable or something.

amulware (approved):
Can't think of any more edge cases (outside of general float weirdness, but what can you do, huh?). The tests don't cover everything perfectly (startpoint/endpoint don't check the exact point (could perhaps have a couple of examples of hardcoded points instead of just duplicating the math?), but are pretty good overall. Consider these nits and then take it away!

Bearded.Utilities.Tests/Generators/DirectionGenerators.cs
tomrijnbeek: Yes, I am not sure if it makes sense to expose the underlying data model, but you can see here that that leads to us converting back and forth. There doesn't seem to be an easy way to easily generate floats in a certain range without just filtering, which may throw away edge cases.

tomrijnbeek: edge cases = test cases

amulware:
Looking fine to me, feel free to merge.

PR #233, opened 2021-09-25
✅ Rewrite circular arc tests to untie from members

tomrijnbeek:
## ✨ What's this?
This PR rewrites the circular arc tests to no longer be grouped by the member they test.

### 🔗 Relationships
N/A

## 🔍 Why do we want this?
We want to focus on behaviours, rather than members in our tests.

The reason we are rewriting an existing test is to have a single canonical example that we can refer contributors to the bearded libraries to as part of our effort to improve test coverage.

## 🏗 How is it done?
This PR attempts to maintain the current test coverage, but rebrand its tests to talk more about the cause and effect, rather than the members it affects. This was largely possible without changing the tests, though some reordering took place to make things make more sense together.

There are sort of two naming schemes in use now:

1. `Cause_Effect`
2. `Fact`

The first is used mostly for the factory methods (e.g. `CreatingShortArc_ReturnsArcWithAngleSmallerThanPiRad`). This is specifically talking about how the arc created in a specific way has certain properties.

The latter is used to describe the properties of the arc (e.g. `OppositeOfAShortArcIsALongArc`), and is primarily used to describe how the different public members of the arc should behave.

I do not know if we should keep these two side by side, whether one is preferred over the other, or whether we don't care at all as long as the name is clear.

### 💥 Breaking changes
N/A

### 🔬 Why not another way?
N/A

### 🦋 Side effects
Once this PR is approved, I will create overarching issues for each bearded library, with the hacktoberfest tag and pinned to the top, referring to these tests as an example of what we'd like to see in further test PRs.

## 💡 Review hints
I am not entirely convinced with the names, they seem really long which harms readability. My recommendation would be to focus on the names, but if you think the tests should be organised/grouped differently, then I keep myself open to feedback.

### DIFF
--- a/Bearded.Utilities.Tests/Geometry/CircularArc2Tests.cs
+++ b/Bearded.Utilities.Tests/Geometry/CircularArc2Tests.cs
@@ -19,7 +19,7 @@ namespace Bearded.Utilities.Tests.Geometry
         }
 
         [Property]
-        public void CreatingShortArcBetweenIdenticalDirections_ReturnsZeroLengthArc(Direction2 fromTo)
+        public void ShortArcBetweenIdenticalDirectionsIsZeroLength(Direction2 fromTo)
         {
             var arc = CircularArc2.ShortArcBetweenDirections(Vector2.Zero, 1, fromTo, fromTo);
 
@@ -27,7 +27,7 @@ namespace Bearded.Utilities.Tests.Geometry
         }
 
         [Property]
-        public void CreatingShortArcBetweenOppositeDirections_ReturnsNegativeDirectionArc(Direction2 from)
+        public void ShortArcBetweenOppositeDirectionsHasMinusPiAngle(Direction2 from)
         {
             var arc = CircularArc2.ShortArcBetweenDirections(Vector2.Zero, 1, from, -from);
 
@@ -35,7 +35,7 @@ namespace Bearded.Utilities.Tests.Geometry
         }
 
         [Property]
-        public void CreatingShortArc_ReturnsArcWithAngleSmallerThanPiRad(Direction2 from, Direction2 to)
+        public void ShortArcHasAngleWithMagnitudeSmallerThanPi(Direction2 from, Direction2 to)
         {
             if (from == to) return;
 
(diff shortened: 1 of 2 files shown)


Bearded.Utilities.Tests/Geometry/CircularArc2Tests.cs
```diff
@@ -10,354 +10,332 @@ namespace Bearded.Utilities.Tests.Geometry
 {
     public sealed class CircularArc2Tests
     {
-        public sealed class ShortArcBetweenDirections
+        private const float epsilon = 0.001f;
+
+        [Property(Arbitrary = new[] { typeof(DirectionGenerators.All) })]
+        public void CreatingShortArcBetweenIdenticalDirections_ReturnsZeroLengthArc(Direction2 fromTo)
```
amulware, line 16: We could rename these into 'facts' like `ShortArcBetweenIdenticalDirectionsIsZeroLength`, which is shorter too (note how I changed "ReturnsZeroLengthArc" to "IsZeroLength"). With that, I don't think the names are too long either, they need that info to be clear.

Bearded.Utilities.Tests/Geometry/CircularArc2Tests.cs
```diff
@@ -10,354 +10,332 @@ namespace Bearded.Utilities.Tests.Geometry
 {
     public sealed class CircularArc2Tests
     {
-        public sealed class ShortArcBetweenDirections
+        private const float epsilon = 0.001f;
+
+        [Property(Arbitrary = new[] { typeof(DirectionGenerators.All) })]
```
amulware, line 15: May be more convenient to register the direction generator that's used everywhere just once in the constructor? (see https://github.com/beardgame/td/blob/develop/src/Bearded.TD.Tests/Game/Generation/AreaTagsTests.cs )

tomrijnbeek: Didn't know about that trick, thanks!

Bearded.Utilities.Tests/Geometry/CircularArc2Tests.cs
tomrijnbeek: Thanks, I like that suggestion. Applied as best as possible throughout all tests.

amulware (approved):

Issue #229, opened 2021-09-24
Implement rotors


tomrijnbeek:
## 🆙 Improvement proposal
Rotors are an alternative to quaternions to implement rotations in 3D (and beyond). I suggest we implement rotors for that reason. My suggestion would be to add types for BiVectors, and implementing to geometric product. From there, implementing rotors should be straightforward.

Note that there is some overlap with #177 to support typesafe rotations. Once we have rotors implemented, they could be used to represent a typesafe rotation (I believe we should keep the mathematical concept of a rotor separate from its geometric purpose).

## Context
https://www.youtube.com/watch?v=60z_hpEAtD8

PR #235, opened 2021-10-05
🟪 Implement bivectors

tomrijnbeek:
## ✨ What's this?
This PR implements representations for bivectors in 2 and 3 dimensions.

### 🔗 Relationships
Required for #229 

## 🔍 Why do we want this?
Bivectors are a geometric construct needed to implement rotors. They are also a basic concept in geometric algebra in general.

## 🏗 How is it done?
The PR adds two new structs: `Bivector2` and `Bivector3` representing bivectors in 2 and 3 dimensions respectively. Note that copies with double precision could be made in case we want to do so.

The additions are accompanied by the necessary tests, including additions to the testing library to make it possible to test these easily. Coverage shows that all the code is 100% covered.

### 💥 Breaking changes
N/A

### 🔬 Why not another way?
Having a type safe way of representing bivectors falls within the general trend in these libraries to be type safe where we can.

The types have been added to the `Geometry` namespace, though I am not convinced that is the right place to put them. I believe that a `Math` or `Mathematics` namespace would be more appropriate. Bivectors have clear geometric uses, but like vectors, can be considered a pure mathematical construct. Remaining consistent with OpenTK, which puts the vector and matrix types in a `Mathematics` namespace, sounds desirable. The reason I avoided adding this namespace is my worry that having a `Math` namespace would cause confusion as to why our math helpers would not be in said namespace, but in `Core`. Making a breaking change is something I wanted to avoid.

Furthermore, some attention needs to be given to the name `OuterProduct`. Outer product isn't the mathematically correct name. "Exterior product" and "wedge product" are more mathematically pure. However, in the context of talking about geometric algebra and rotors, the term "outer product" is used, and so Googling for this concept may yield better results. I can be convinced to rename though. My preference would be `Bivectorn.Wedge` in that case.

### 🦋 Side effects
Further code changes will be needed to use these bivectors in rotors.

## 💡 Review hints
Please comment on the open questions in "why not another way?" section above.

Further reading to understand the subject matter:

* 📚 [Let's remove quaternions from every engine](https://marctenbosch.com/quaternions/)
* 💻 [Reference implementation by Marc ten Bosch](https://marctenbosch.com/quaternions/code.htm)
* 📚 [Bivector Wikipedia article](https://en.wikipedia.org/wiki/Bivector)
* 🎥 [A Swift Introduction to Geometric Algebra](https://www.youtube.com/watch?v=60z_hpEAtD8)

### DIFF
--- a/Bearded.Utilities.Testing/Geometry/Bivector2Assertions.cs
+++ b/Bearded.Utilities.Testing/Geometry/Bivector2Assertions.cs
@@ -1,0 +1,42 @@
+using Bearded.Utilities.Geometry;
+using FluentAssertions;
+
+namespace Bearded.Utilities.Testing.Geometry
+{
+    public sealed class Bivector2Assertions
+    {
+        private readonly Bivector2 subject;
+
+        public Bivector2Assertions(Bivector2 instance) => subject = instance;
+
+        [CustomAssertion]
+        public AndConstraint<Bivector2Assertions> Be(Bivector2 other, string because = "", params object[] becauseArgs)
+        {
+            AssertionExtensions.Should(subject).Be(other, because, becauseArgs);
+            return new AndConstraint<Bivector2Assertions>(this);
+        }
+
+        [CustomAssertion]
+        public AndConstraint<Bivector2Assertions> NotBe(Bivector2 other, string because = "", params object[] becauseArgs)
+        {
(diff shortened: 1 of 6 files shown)


tomrijnbeek:
This PR is now ready to review.

Note that this implementation is very much intended as a minimum viable product, and may be further expanded upon as needed by rotors down the line.

amulware:
Feel free to merge if you like :)

PR #238, opened 2021-10-09
🔥 Delete Mathf

### LINKED ISSUE #201: Deprecate Mathf in favour of System MathF

tomrijnbeek:
## ✨ What's this?
This deletes the class `Mathf`, which has been obsolete for some time.

### 🔗 Relationships
Closes #201 

## 🔍 Why do we want this?
There is now a `System.MathF` class with all this functionality. Having only one way to do things is good.

## 🏗 How is it done?
Select the file, hit "delete".

### 💥 Breaking changes
Removes the `Mathf` class, so any dependencies on this class will break.

## 💡 Review hints
Feel free to squash and merge.


amulware:
I did not realise this existed since core 2.0!

PR #245, opened 2021-10-21
Destroy_Game_Object

CursedMuffins:
A simple code that helps with the destroying of object once it touched the player, this is useful in mini games.

### 🔗 Relationships

Closes #

### DIFF
--- a/Destroy_Game_Object
+++ b/Destroy_Game_Object
@@ -1,0 +1,12 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Enemy_Destroy : MonoBehaviour
+{
+    Rigidbody rb;
+    private void OnCollisionEnter(Collision collision)
+    {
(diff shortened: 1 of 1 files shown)


amulware:
@CursedMuffins Your last three PRs here look like they are about adding Unity components. Unfortunately:
- This is not a Unity library (though it can certainly be used with Unity) and we have no intention of adding that dependency
- You added the code in unusually named files without extensions at the root of the repo (instead of inside one of the projects) 
- You gave very little information in the PR description of why we would want to add this, and did not refer to any issues (as a third-party contributor it's usually better to find an existing issue or at least propose your own before making changes, to make sure your efforts are in line with the maintainers vision for the repo)

Especially within the current context of Hacktoberfest, these PRs look a lot like intentional spam. Please be a little more careful and considerate when contributing to open source in the future. There are many good resources on how to get started!

PR #251, opened 2021-12-11
⏮ Switch to file scope namespaces

paulcscharf:
## ✨ What's this?
What it says on the tin. Rewriting pretty much all files to remove namespace blocks in favour of file scope namespaces.

### 🔗 Relationships
- [x] ~~Depends on #250, which needs to be merged first.~~

## 🔍 Why do we want this?
We don't need it, but we like it. I think biting the bullet all at once is better than to change files overtime, especially if we're making other changes at the same time (which we are likely to, especially with how we like to squash PRs).


paulcscharf:
Force pushed since the PR this depended one was already merged and this keeps things a bit cleaner.

tomrijnbeek (approved):

PR #300, opened 2022-10-24
Transition to using explicit nullable types (#184)

### LINKED ISSUE #184: Enforce nullable reference types

artog:
## ✨ What's this?
Removes all warnings related to nullable. Also made a more straight-forward implementation of `Maybe<T>`

### 🔗 Relationships
Closes #184 

## 🔍 Why do we want this?
Using explicit nullable is a move towards safer and more stable code.

## 🏗 How is it done?
Mostly went through all errors and fixes as I went along. `Maybe<T>` caused a bunch of trouble, so I remade the functionallity using an abstract main class (`Maybe<T>`) with two deriving classes (`Just<T>` and `Nothing<T>`). This separates the intent into two different classes, both simplifying the actual logic and prevents any casting/default issues.

### 💥 Breaking changes
Hopefully none. Tests are unchanged (and still full of warnings) to allow for insight into backwards compability.

## 💡 Review hints
Might be wasier to look at commits rather than full PR, especially for the change to `Maybe<T>`

### DIFF
--- a/Bearded.Utilities/Algorithms/CoffmanGraham.cs
+++ b/Bearded.Utilities/Algorithms/CoffmanGraham.cs
@@ -177,7 +177,7 @@ public static class CoffmanGraham
 
     public static ISolver SolverForReducedGraphs(int maxLayerSize) => new ReducedGraphSolver(maxLayerSize);
 
-    private struct DecreasingNumberSequence : IComparable<DecreasingNumberSequence>, IComparable
+    private readonly struct DecreasingNumberSequence : IComparable<DecreasingNumberSequence>, IComparable
     {
         private readonly ImmutableList<int> numbers;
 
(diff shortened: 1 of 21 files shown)


artog:
Completed transition for `Bearded.Utilities`, next up is `Bearded.Testing`.

tomrijnbeek:
Hi artog,

Great to see you've taken an interest in contributing. The nullability issue is one that would be great to close off finally!

I didn't review your PR in detail, since it's still in draft, but the change you made to `Maybe` caught my eye and I wanted to briefly chime in before you go down a rabbit hole there. A complete reimplementation of `Maybe` is something I would consider out of scope of just the nullability change. Our implementation is specifically meant as a performance-friendly implementation, which is why we chose for it to be a struct, and not a class. The proposed implementation undermines those principles, so it sounds unlikely that we would adopt this into the library (since there are other implementations that follow the same pattern already).

The `Maybe` type was built in the time before nullable became widely adopted, so I think its value in existing is debatable in the first place.

My suggestion would be to either revert the `Maybe` changes, or split them into a separate PR if you feel like you have a strong case for those changes.

PR #317, opened 2023-04-26
📝 Add NuGet metadata

Mattias1:
## ✨ What's this?
Add some metadata for the NuGet gallery (https://www.nuget.org/packages/Bearded.Utilities).
Specifically, a link back to the github repo, a MIT licence and a good description.

## 🔍 Why do we want this?
More clarity for people using the NuGet gallery.

## 🏗 How is it done?
Mostly moving metadata from the (no longer used) nuspec file to the project file, as is the best practice these days (https://learn.microsoft.com/en-us/nuget/create-packages/package-authoring-best-practices).

### 🔬 Why not another way?
It's best practice these days to use the project file instead of nuspec files (which you weren't using anymore anyways).

### 🦋 Side effects
Maybe the tags aren't up to date anymore - I left out most tags from the repository, as they seem a bit overly broad.


Mattias1:
This should probably get a tag "* `ignore-for-release`"
How to add this?

tomrijnbeek:
Should be under labels:

image

Though unsure if that is only available for contributors.

Mattias1:
Ah, yeah, I don't think I can add labels in your repository.

tomrijnbeek:
Weird, will add them.

tomrijnbeek:
Benchmarks probably should not be pushed to NuGet, that's a missing flag to exclude that from publishing I think. We could also really be improving the descriptions, but these are the same as from the nuspec, so all in all, this is a definite improvement. Thanks.

